Issue #42, opened 2014-10-10
Add special handling to images that have geometry inconsistencies


fedorov:
Need to improve handling of the following 2 cases:

1. GUI usage of Reporting module, when an input label is selected, and its geometry (spacing, orientation, dimensions) do not match that of the input image. User should be given a warning about this situation, and should be given an option to create a new label (by resampling) that would match the geometry of the image.

2. For conversions of the label images using the command line script, warning should be printed on the console, but resampling should probably take place by default (https://github.com/fedorov/Reporting/blob/master/Util/LabelToDICOMSEGConverterScript.py).

@naucoin let's talk about this next time we meet.

naucoin:
Updates made to Slicer to avoid the false positives in the geometry check:
https://github.com/Slicer/Slicer/commit/9fe70aaf70e7e24081c20e7eff7fba236eed5108

pieper:
Nice!

fedorov:
Thanks Nicole! Are all the pieces in the nightly so we can check with Jayashree that she is satisfied with the result?

naucoin:
The pieces are in the Slicer nightly, but the pull request for Reporting hasn't been integrated yet: https://github.com/fedorov/Reporting/pull/45

fedorov:
Are you waiting for me to review and merge it?

naucoin:
@fedorov Yes

fedorov:
Ok, I didn't realize that. You merged some pull requests in the past, so I thought you would do the same here. I will review and comment/merge.

naucoin:
I was more waiting on the review, sorry I wasn't clear, I wanted to be sure I hit all the to-do points for this issue.

naucoin:
I updated my github branch to use vtkImageCast and to test the converter script. I ended up having to take out the try block in order to debug it, and checked it in like that. I think that it's more important to know why the script failed since any exceptions were treated generically ("An error has occurred" was the only print out). @fedorov if you don't agree, I can put it back in.

naucoin:
Also: it's only ensuring the cast is done after the resample, do you want to always check and cast the label map to unsigned short?

fedorov:
> I updated my github branch to use vtkImageCast and to test the converter script. I ended up having to 
> take out the try block in order to debug it, and checked it in like that.

that's fine

> Also: it's only ensuring the cast is done after the resample, do you want to always check and cast the 
> label map to unsigned short?

[...]

PR #45, opened 2014-12-31
42 rebase add special handling to images that have geometric inconsistencies

naucoin:
BUG: recover from selecting segmentation with wrong geometry
    
Offer to the user the option of resampling the chosen segmentation,
preserving any labels in it, to create a new label volume for the segmentation.
Added a new helper class for a popup question, returns True or False for a Yes or No
answer.

Call the volumes logic method CheckForLabelVolumeValidity to get a
verbose message string if the label map and scalar volumes don't
match geometry wise. An empty string means that they match, otherwise
print it out in the error message box.

Remove GeometriesMatch since we're now using the more extensive volumes
logic volumesLogic.CheckForLabelVolumeValidity
  
Update script to check geometry before resampling, add a geometry check and
only resample if necessary. Still use the BRAINS resample, the volumes logic
resample doesn't enforce the ushort output type.

Issue #42


fedorov:
Nicole, looks good to me!

One suggestion: I think it makes sense to change the converter script to use the same resampling mechanism as done in the module (right now the script is using BRAINS, and the module is using Volumes logic). This way we (hopefully) make sure there is no incidental inconsistency while converting the label using GUI and the script, and also make the code more symmetric. 

How about we just cast the input label to ushort in the converter script, and then call Volumes resample? Or is there something else I didn't think of?

naucoin:
I double checked the volumes logic resample, it will use nearest neighbour interpolation if the input volume is a label map, and that flag is already set. 
As long as the input volume is ushort, the output will be as well. The label volume is created as a new vtkMRMLScalarVolumeNode that's been flagged as a label volume after reading in the data from disk. Doesn't look like in the node that it triggers any re-sampling, so we'll have to do that before passing it in.

naucoin:
Committed the change to use the volumes logic resampling and volumes info widget type conversion.

naucoin:
Thanks! I'll see if I can get the dashboard cleaned up as well.

fedorov:
Nicole, are you planning to update ExtensionIndex?

naucoin:
Extensions index pull request:
https://github.com/Slicer/ExtensionsIndex/pull/936

naucoin:
A couple of tests are failing due to the logging change in Reporting:
https://github.com/fedorov/Reporting/commit/7b99e96fc1374cf5c972dab78dd5f6de793fa15e
http://slicer.cdash.org/testDetails.php?test=6164083&build=410553

fedorov:
Thanks!

naucoin:
ReportingRoundTripTest: after adding the import logging call, digging into why the fiducial is at the wrong place: when the DICOM volume gets loaded back in from the report, it's IJK2RAS matrix is identity, instead of what it was when it was originally loaded. Therefore the conversion of the fiducial from the slice's display coordinates to RAS is failing. Digging into why the matrix is missing on reload.
Also: have to update to Markups fiducials at some point.

PR #55, opened 2015-11-04
Refactor to new dcmtk

pieper:
WIP - merge if you want.  This finishes the work of putting the seg CLIs into reporting.


pieper:
I see this also includes the new dcmtk stuff so maybe you don't want to merge this directly.  Close it if you aren't ready for that to be in master.

SEGSupport/CMakeLists.txt
```diff
@@ -1,29 +1,42 @@
...
-  framesorter.h
-  SegmentAttributes.h
-  )
-
+#-----------------------------------------------------------------------------
 set(MODULE_TARGET_LIBRARIES
+  ${DCMTK_LIBRARIES}
   ${ITK_LIBRARIES}
-  )
+  xml2 z
```
jcfr: Why are these libraries needed here ?

SEGSupport/CMakeLists.txt
```diff
@@ -1,29 +1,42 @@
...
-  ADDITIONAL_SRCS ${MODULE_SRCS}
-  )
+  ADDITIONAL_SRCS EncodeSEG.cxx
+  TARGET_LIBRARIES ${MODULE_TARGET_LIBRARIES}
+  RUNTIME_OUTPUT_DIRECTORY        "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
+  LIBRARY_OUTPUT_DIRECTORY        "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}"
+  ARCHIVE_OUTPUT_DIRECTORY        "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}"
+  INSTALL_RUNTIME_DESTINATION ${CMAKE_BINARY_DIR}
+  INSTALL_LIBRARY_DESTINATION ${CMAKE_BINARY_DIR}
+  INSTALL_ARCHIVE_DESTINATION ${CMAKE_BINARY_DIR}
```
jcfr: Why is it needed to explicitly set these ? 

```
RUNTIME_OUTPUT_DIRECTORY        "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}"
+  LIBRARY_OUTPUT_DIRECTORY        "${CMAKE_LIBRARY_OUTPUT_DIRECTORY}"
+  ARCHIVE_OUTPUT_DIRECTORY        "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}"
+  INSTALL_RUNTIME_DESTINATION ${CMAKE_BINARY_DIR}
+  INSTALL_LIBRARY_DESTINATION ${CMAKE_BINARY_DIR}
+  INSTALL_ARCHIVE_DESTINATION ${CMAKE_BINARY_DIR}
```

If you configure the SlicerExecutionModel with different default, they will be installed and built in the location of your choice.

By forcing these values .. it will basically break the extension packaging.

SEGSupport/EncodeSEG.cxx
```diff
@@ -38,12 +38,12 @@
...
 // UIDs
-#include "../Common/QIICRUIDs.h"
-#include "../Common/SegmentAttributes.h"
+#include "QIICRUIDs.h"
+#include "SegmentAttributes.h"
 //#include "../Common/conditionCheckMacros.h"
 
 // versioning
-#include "../Iowa2DICOMVersionConfigure.h"
+#include "../vtkReportingVersionConfigure.h"
```
jcfr: Would be nice to avoid relative important. If needed, the directory should be added as `INCLUDE_DIRECTORIES` when calling `SEMMacroBuildCLI`

SEGSupport/CMakeLists.txt
pieper: Not sure - @fedorov had them in the https://github.com/QIICR/Iowa2DICOM repository so I brought them over.

fedorov: I had some compile problems, but I think those might have been resolved after DCMTK cmake issues were resolved. Need to test. Steve, did you try building without those?

SEGSupport/EncodeSEG.cxx
pieper: The relative path here seems more readable and explicit - otherwise the info is hidden away and it's not obvious where to look for this (already somewhat magical) file.

SEGSupport/CMakeLists.txt
jcfr: I see .. and the original commit is not helpful either. 
See QIICR/Iowa2DICOM@8302cae28c

@fedorov If this is an issue with building DCMTK, would be great to document it ... so that we can fix it. :smile:

SEGSupport/CMakeLists.txt
fedorov: @jcfr I think the issue is that not all developers know how this macro is actually working, and just follow examples from other places (at least that is what I do).

I checked another extension, and guess we need to update to follow that pattern instead: https://github.com/fedorov/ChangeTrackerPy/blob/master/Metrics/IntensityDifference/CMakeLists.txt#L13-L18

SEGSupport/EncodeSEG.cxx
fedorov: The "somewhat magical" file is cmake-generated at build time.

SEGSupport/CMakeLists.txt
fedorov: I don't think this is an issue anymore - the other CLI compiles fine without those libraries. I think those libs should be removed.

SEGSupport/EncodeSEG.cxx
naucoin: Would it be picked up by having the binary directory included in the include paths?

pieper:
This latest version builds for me without the extra cmake directives and without the extra libraries.

SEGSupport/EncodeSEG.cxx
pieper: Nicole is right, we typically do include the build path in the list of include paths.

fedorov:
Steve, will you update the PR to incorporate all the feedback above?

pieper:
I believe all the comments have been included in the most recent version.

fedorov:
Thanks!

Issue #52, opened 2015-09-03
Review ClearCanvas plugin terminology for reuse


fedorov:
See 

Segmentation/Configuration/AnatomicRegionAndModifier.xml
Segmentation/Configuration/SegmentationCategoryTypeModifier.xml

in https://github.com/fedorov/ClearCanvas-AimPlugin4.5

fedorov:
https://github.com/fedorov/ClearCanvas-AimPlugin4.5/blob/master/Segmentation/Configuration/AnatomicRegionAndModifier.xml
https://github.com/fedorov/ClearCanvas-AimPlugin4.5/blob/master/Segmentation/Configuration/SegmentationCategoryTypeModifier.xml

fedorov:
To clarify this issue, anatomic region and segmentation category need to be populated in DICOM SEG, see here as an example: https://github.com/QIICR/Iowa2DICOM/blob/master/ConvertSegmentations/EncodeSEG.cxx#L346-L352

The idea is to have a free text dynamic search based on code meaning for the the region and modifier, and for category and its modifier. It is a quite long list, and I don't know how dynamic search is implemented in Qt or other tools we have. That needs to be investigated.

Let me know if we need to discuss more.

naucoin:
The Models module has a "scroll to" search box, and the color picker pop up also has a search box, so there are some examples I can follow.

fedorov:
The module search box in the mail toolbar is also a good one.

naucoin:
@fedorov: after our meeting today, my understanding of the phases of this change are:

1. Add information (non-editable) to the Editor Module under the color picker Label widgets that show the color name, value and color to show the terminology that has been looked up:
Category: [       ] Modifier: [     ]
Region: [    ]   Modifier: [     ]
When the user picks a color, the informatin is updated, but the boxes won't show if the color table isn't linked to terminology (or should we show generic values?). Add this information to the label map via node attributes.

2. Allow the user to edit the Category and Region by clicking on them to search a list of terms. This will allow resetting from the generic values.

3. If a user has chosen a color table that doesn't have the terminology mapping defined, create a way to edit them easily for each value (add columns to the Colors module editor? workflow that gives you one label at a time?). Save the new terminology mapping with the segmentation object.

To keep in mind:
- terminology may be on a remote server
- further integration of the terminology widget across Slicer
- coordinate with the Queen's group on the new segmentation and editor work

fedorov:
[...]

PR #57, opened 2015-11-06
ENH: Add terminology region support, fix color display

naucoin:
After updates to the Slicer topic branch in naucoin
4047-Add-semantic-color-information-to-color-selector-in-editor
parse out and set the region information read from the segmentation object. Take
into account that the modifier lines may be omitted in the .info file, refer
to SEG2NRRD for the calls that do the writing.
Updated SEG2NRRD to set the label map pixel values to the segmentation number
rather than always using 1. This allows the colors to display properly with one
color node rather than having to define a color node per segmentation.

Issue #52

### DIFF
--- a/Py/DICOMSegmentationPlugin.py
+++ b/Py/DICOMSegmentationPlugin.py
@@ -143,54 +143,67 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
       (success,labelNode) = slicer.util.loadLabelVolume(labelFileName, returnNode=True)
 
       # Initialize color and terminology from .info file
-      # Format of the .info file:
+      # See SEG2NRRD.cxx for how it's written.
+      # Format of the .info file (no leading spaces, modifiers are optionally written if present):
       #    RGBColor:128,174,128
       #    AnatomicRegion:T-C5300,SRT,pharyngeal tonsil (adenoid)
+      #    AnatomicRegionModifier:code,scheme,meaning
       #    SegmentedPropertyCategory:M-01000,SRT,Morphologically Altered Structure
-      #     SegmentedPropertyType:M-80003,SRT,Neoplasm, Primary
+      #    SegmentedPropertyType:M-80003,SRT,Neoplasm, Primary
+      #    SegmentedPropertyTypeModifier:code,scheme,meaning
+      colorIndex = segmentId + 1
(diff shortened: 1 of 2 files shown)


naucoin:
Note: ready to merge, up to date with my current Slicer branch 
https://github.com/naucoin/Slicer/tree/4047-Add-semantic-color-information-to-color-selector-in-editor

Py/DICOMSegmentationPlugin.py
```diff
@@ -143,54 +143,67 @@ def load(self,loadable):
       (success,labelNode) = slicer.util.loadLabelVolume(labelFileName, returnNode=True)
 
       # Initialize color and terminology from .info file
-      # Format of the .info file:
+      # See SEG2NRRD.cxx for how it's written.
+      # Format of the .info file (no leading spaces, modifiers are optionally written if present):
```
fedorov: Not only modifiers, but complete attributes may be missing. It probably makes sense to set Category=Tissue Type=Tissue of none is present. Anatomic Region is optional. Category and Type are mandated by the standard, but we should not crash if the object is non-compliant.

naucoin: Makes sense, I'll initialize all the strings. 
In future if you've got feedback on a pull request that requires changes, it would be best if you held off on merging as it's easier for me to add another commit to the topic branch and squash them once I've addressed all your concerns.

fedorov: The reason I merged is that I don't want to spend time refining all the wrinkles in the first iteration. "Release often release early". Error checking will need to be refined anyway at different levels, including SEG2NRRD.

naucoin: One round of unwrinkling would be useful as there's more over head on my side to get the minor change in. I've updated the file on this branch to init the strings, now looking at how to get that into your branch (feel free to cherry-pick it):
https://github.com/naucoin/Reporting/commit/35a789c6504e8c4ddcef46b0904a08e723b5ae4d

fedorov: I will. At the moment, working on another extension that also needs to go to Slicer asap.

fedorov:
Done, I could do it with the web GUI!

naucoin:
Great!

PR #66, opened 2015-11-17
Add CIE color to the encoding process

pieper:


fedorov:
I guess you meant RGB color, right?

pieper:
Well, I meant adding CIE color to the resulting segmentation object by adding RGB to the EncodeSEG parameters.

PR #68, opened 2015-11-23
Integrate editor seg export

pieper:
This matches the pull request with the same name in the Slicer trunk.

I will push the Slicer one to svn (it should be okay even if this doesn't get updated in sync).


fedorov:
:+1: this is great, thanks!

I am going to merge, since I don't have time to carefully review at the moment, but it looks good on the quick look, and I will plan to test together with the updated Slicer later tonight.

PR #71, opened 2016-01-10
Add support for the Segmentations Module

Sunderlandkyl:
DicomSegmentation.py
  * Added code within the import functions that adds the DICOM segmentation to a segmentation object.
  * Added functions: examineForExport, export, and exportAsDICOMSEG (which is a modified version of exportAsDICOMSEG from EditUtil.py in the Slicer core.

SEG2NRRD.cxx
  * Changes made by Andras that prevents the CLI from crashing due to an incorrectly allocated number of Slices.

### DIFF
--- a/Py/DICOMSegmentationPlugin.py
+++ b/Py/DICOMSegmentationPlugin.py
@@ -351,6 +351,17 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
     exportable = None
 
     if node.GetAssociatedNode() and node.GetAssociatedNode().IsA('vtkMRMLSegmentationNode'):
+
+      # Check to make sure all referenced UIDs exist in the database.
+      instanceUIDs = node.GetAttribute("DICOM.ReferencedInstanceUIDs").split()
+      if instanceUIDs == "":
+          return []
+
+      for instanceUID in instanceUIDs:
+        inputDICOMImageFileName = slicer.dicomDatabase.fileForInstance(instanceUID)
+        if inputDICOMImageFileName == "":
+          return []
+
       exportable = slicer.qSlicerDICOMExportable()
       exportable.confidence = 1.0
       exportable.setTag('Modality', 'SEG')
--- a/Py/DICOMSegmentationPlugin.py
+++ b/Py/DICOMSegmentationPlugin.py
@@ -252,6 +252,9 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
         # end of processing a line of terminology
       infoFile.close()
 
+      #TODO: Create logic class that both CLI and this plugin uses so that we don't need to have temporary NRRD files and labelmap nodes
+      #if not hasattr(slicer.modules, 'segmentations'):
+
       # load the segmentation volume file and name it for the reference series and segment color
       labelFileName = os.path.join(outputDir,str(segmentId+1)+".nrrd")
       segmentName = seriesName + "-" + colorName + "-label"
@@ -294,6 +297,51 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
     segmentationColorNode.NamesInitialisedOn()
 
     # TODO: the outputDir should be cleaned up
+
+    if hasattr(slicer.modules, 'segmentations'):
+
+      import vtkSlicerSegmentationsModuleLogic
+      import vtkSlicerSegmentationsModuleMRML
+      import vtkSegmentationCore
+
+      segmentationNode = vtkSlicerSegmentationsModuleMRML.vtkMRMLSegmentationNode()
+      segmentationNode.SetName(seriesName)
+      slicer.mrmlScene.AddNode(segmentationNode)
+
+      segmentationDisplayNode = vtkSlicerSegmentationsModuleMRML.vtkMRMLSegmentationDisplayNode()
+      segmentationNode.SetAndObserveDisplayNodeID(segmentationDisplayNode.GetID())
+      slicer.mrmlScene.AddNode(segmentationDisplayNode)
+
+      segmentation = vtkSegmentationCore.vtkSegmentation()
+      segmentation.SetMasterRepresentationName(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationBinaryLabelmapRepresentationName())
+
+      segmentationNode.SetAndObserveSegmentation(segmentation)
+      self.addSeriesInSubjectHierarchy(loadable, segmentationNode)
+
+      colorID = 1
+      for segmentNode in segmentNodes:
+        segment = vtkSegmentationCore.vtkSegment()
+        segment.SetName(segmentNode.GetName())
+
+        segmentColor = [0,0,0,0]
+        segmentationColorNode.GetColor(colorID, segmentColor)
+        segment.SetDefaultColor(segmentColor[0:3])
+
+        colorID += 1
+
+        #TODO: when the logic class is created, this will need to be changed
+        logic = vtkSlicerSegmentationsModuleLogic.vtkSlicerSegmentationsModuleLogic()
+        orientedImage = logic.CreateOrientedImageDataFromVolumeNode(segmentNode)
+        segment.AddRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationBinaryLabelmapRepresentationName(), orientedImage)
+        segmentation.AddSegment(segment)
+
+        segmentDisplayNode = segmentNode.GetDisplayNode()
+        slicer.mrmlScene.RemoveNode(segmentDisplayNode)
+        slicer.mrmlScene.RemoveNode(segmentNode)
+
+      segmentation.CreateRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationClosedSurfaceRepresentationName(), True)
+      slicer.mrmlScene.RemoveNode(segmentationColorNode)
+      slicer.mrmlScene.RemoveNode(mergeNode)
 
     return True
 
--- a/Py/DICOMSegmentationPlugin.py
+++ b/Py/DICOMSegmentationPlugin.py
@@ -386,155 +386,157 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
 
     import logging
     
-    exportable = exportablesCollection.GetItemAsObject(0)
-    subjectHierarchyNode = slicer.mrmlScene.GetNodeByID(exportable.GetNodeID())
-
-    instanceUIDs = subjectHierarchyNode.GetAttribute("DICOM.ReferencedInstanceUIDs").split()
-
-    if instanceUIDs == "":
-      raise Exception("Editor master node does not have DICOM information")
-
-    # get the list of source DICOM files
-    inputDICOMImageFileNames = ""
-    for instanceUID in instanceUIDs:
-      inputDICOMImageFileNames += slicer.dicomDatabase.fileForInstance(instanceUID) + ","
-    inputDICOMImageFileNames = inputDICOMImageFileNames[:-1] # strip last comma
-
-    # save the per-structure volumes in the temp directory
-    inputSegmentationsFileNames = ""
-    import random # TODO: better way to generate temp file names?
-    import vtkITK
-
-    writer = vtkITK.vtkITKImageWriter()
-    rasToIJKMatrix = vtk.vtkMatrix4x4()
-
-    import vtkSegmentationCore
-    import vtkSlicerSegmentationsModuleLogic
-    logic = vtkSlicerSegmentationsModuleLogic.vtkSlicerSegmentationsModuleLogic()
-
-    segmentationTransform = vtk.vtkMatrix4x4()
-    segmentationNode = subjectHierarchyNode.GetAssociatedNode()
-
-    mergedSegmentationImageData = segmentationNode.GetImageData()
-    mergedSegmentationLabelmapNode = slicer.vtkMRMLLabelMapVolumeNode()
+    if hasattr(slicer.modules, 'segmentations'):
     
-    segmentationNode.GetRASToIJKMatrix(rasToIJKMatrix)
-    mergedSegmentationLabelmapNode.SetRASToIJKMatrix(rasToIJKMatrix)
-    mergedSegmentationLabelmapNode.SetAndObserveImageData(mergedSegmentationImageData)
-    mergedSegmentationOrientedImageData = logic.CreateOrientedImageDataFromVolumeNode(mergedSegmentationLabelmapNode)
+      exportable = exportablesCollection.GetItemAsObject(0)
+      subjectHierarchyNode = slicer.mrmlScene.GetNodeByID(exportable.GetNodeID())
 
-    segmentation = segmentationNode.GetSegmentation()
+      instanceUIDs = subjectHierarchyNode.GetAttribute("DICOM.ReferencedInstanceUIDs").split()
 
-    segmentIDs = vtk.vtkStringArray()
-    segmentation.GetSegmentIDs(segmentIDs)
-    segmentationName = segmentationNode.GetName()
+      if instanceUIDs == "":
+          raise Exception("Editor master node does not have DICOM information")
 
-    for i in range(0, segmentIDs.GetNumberOfValues()):
-      segmentID = segmentIDs.GetValue(i)
-      segment = segmentation.GetSegment(segmentID)
+      # get the list of source DICOM files
+      inputDICOMImageFileNames = ""
+      for instanceUID in instanceUIDs:
+        inputDICOMImageFileNames += slicer.dicomDatabase.fileForInstance(instanceUID) + ","
+      inputDICOMImageFileNames = inputDICOMImageFileNames[:-1] # strip last comma
 
-      segmentName = segment.GetName()
-      structureName = segmentName[len(segmentationName)+1:-1*len('-label')]
+      # save the per-structure volumes in the temp directory
+      inputSegmentationsFileNames = ""
 
-      structureFileName = structureName + str(random.randint(0,vtk.VTK_INT_MAX)) + ".nrrd"
-      filePath = os.path.join(slicer.app.temporaryPath, structureFileName)
-      writer.SetFileName(filePath)
+      import random # TODO: better way to generate temp file names?
+      import vtkITK
+      writer = vtkITK.vtkITKImageWriter()
+      rasToIJKMatrix = vtk.vtkMatrix4x4()
 
-      segmentImageData = segment.GetRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationBinaryLabelmapRepresentationName())
-      paddedImageData = vtkSegmentationCore.vtkOrientedImageData()
-      vtkSegmentationCore.vtkOrientedImageDataResample.PadImageToContainImage(segmentImageData, mergedSegmentationOrientedImageData, paddedImageData)
+      import vtkSegmentationCore
+      import vtkSlicerSegmentationsModuleLogic
+      logic = vtkSlicerSegmentationsModuleLogic.vtkSlicerSegmentationsModuleLogic()
 
-      labelmapImageData = slicer.vtkMRMLLabelMapVolumeNode()
+      segmentationTransform = vtk.vtkMatrix4x4()
+      segmentationNode = subjectHierarchyNode.GetAssociatedNode()
 
-      logic.CreateLabelmapVolumeFromOrientedImageData(paddedImageData, labelmapImageData)
+      mergedSegmentationImageData = segmentationNode.GetImageData()
+      mergedSegmentationLabelmapNode = slicer.vtkMRMLLabelMapVolumeNode()
 
-      writer.SetInputDataObject(labelmapImageData.GetImageData())
+      segmentationNode.GetRASToIJKMatrix(rasToIJKMatrix)
+      mergedSegmentationLabelmapNode.SetRASToIJKMatrix(rasToIJKMatrix)
+      mergedSegmentationLabelmapNode.SetAndObserveImageData(mergedSegmentationImageData)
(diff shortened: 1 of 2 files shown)


Py/DICOMSegmentationPlugin.py
```diff
@@ -295,8 +299,276 @@ def load(self,loadable):
...
+        segmentDisplayNode = segmentNode.GetDisplayNode()
+        slicer.mrmlScene.RemoveNode(segmentDisplayNode)
+        slicer.mrmlScene.RemoveNode(segmentNode)
+
+      segmentation.CreateRepresentation(vtkSegmentationCore.vtkSegmentationConverter.GetSegmentationClosedSurfaceRepresentationName(), True)
+      slicer.mrmlScene.RemoveNode(mergeNode)
+
     return True
 
+  def examineForExport(self, node):
```
fedorov: somewhere here should be a check that the image volume this segmentation corresponds to has DICOM instances, otherwise segmentation would not be exportable into DICOM SEG (at least, not with this plugin).

fedorov:
Kyle, looks fine to me, just one comment that I left. Does it make sense, can you address that?

Sunderlandkyl:
I think so. Does my latest commit properly address the issue?

fedorov:
I see your point, but this is not how I was thinking about this. But maybe this is indeed sufficient. The question is - will DICOM.ReferencedInstanceUIDs that you are checking be initialized somewhere for the segmentations that are created from scratch? I know they will be initialized for segmentations loaded from DICOM.

Sunderlandkyl:
No, the ReferencedInstanceUIDs are not initialized in segmentations created from scratch.
Do you have any recommendations on what the correct behavior would be?

fedorov:
Hopefully, we know through the subject hierarchy the imaging volume that was segmented in a given segmentation object. If this is the case, you should be able to just check the DICOM.InstanceUIDs attribute for that volume. Of course, it becomes more complicated for Segmentations as opposed to traditional Slicer label maps, since you have various representations and the segmentation may correspond to a subset of slices.

fedorov:
Kyle, I am going to merge this, since I think what you did is already very helpful, but it would be great if you could address this remaining issue.

cpinter:
Hi Andriy,
I'm trying to understand the issue here, and it may be more complex than some simple checks. If I understand correctly, the requirement is to save the referenced instance UIDs in the exported segmentation and make sure that a volume exists with those UIDs (not sure if you mean UIDs from nodes or the actual DICOM files). Is that correct?
The problem is that the connection between the master volume and the segmentation is not made currently if you create a segmentation from scratch. It's because the only way to do that until the segmentation editor is created is to import labelmaps or models from nodes, and during importing we do not propagate the connection. I believe Editor references the master volume in a way, maybe we could try to propagate that. But then if the master volume was not loaded from DICOM and do not have UIDs, then we have no UIDs to reference. In RT we solve this by exporting the master volume and the structure set together, and create the UID connections during the joint export. We could create 2.hash random UIDs, but I think it should be a more general mechanism, because then we would need to generate UIDs for all non-DICOM volumes.
That said, I don't understand exactly what the CLI needs for export, so it may be a simpler issue. What do you think would satisfy the requirements?

cpinter:
Also do you see the separation of the CLI to a logic class and a thinner CLI happening in the near future? It would really make this much cleaner on the Slicer side.

pieper:
Hi Csaba -

As I understand the SEG object it can be created without reference to the
instances it segments or with multiple references [1].  Of course they must
be proper DICOM and typically that's the only place to get the needed
metadata so the current Editor implementation requires that the master be a
DICOM or it will raise and exception [2].  I think this is the only
reasonable behavior for now, but in general you might allow people to make
the associations after the fact, like in the subject hierarchy.

I agree it would be ideal if the underlying classes of the SEG related CLIs
could be exposed in a logic class.

-Steve

[1]
http://dicom.nema.org/medical/dicom/current/output/html/part03.html#sect_A.51

[2]
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/EditorLib/EditUtil.py#L372

fedorov:
Yes, referenced instance UIDs is optional. The current CLI requires them, but they don't need to be listed in the ReferencedInstanceUIDs sequence if the image geometry is fully defined in the functional groups, which it is right now. 

However, if you don't have master DICOMs, populating a valid/meaningful DICOM SEG is hard, because all of the equipment/study/patient context has to be "invented", and we cannot use DCMTK functions that carry those forward into the derived object. So it is not populating ReferencedInstanceUIDs is not at all the only/main reason to pass source image DICOM files. All in all, I think a better way to handle this is the same as Csaba you described for RT - if the master volume was not exported from DICOM, create its DICOM representation first. From the egoistic QIICR perspective, if your input image is not in DICOM, it is outside QIICR use case.

Yes, I agree it would be nice to have logic classes, but I thought we discussed the various pros and cons in this thread http://slicer-devel.65872.n3.nabble.com/Converting-ITK-image-to-Slicer-ScalarVolumeNode-td4035458.html#a4035483, and in the end agreed CLI is the most optimal way to go.

Abstracting into a Slicer logic raises many questions: 
* ITK to/from VTK image
* where in Slicer it would fit
* extra dependencies for CLI on libraries other than ITK and DCMTK; I would prefer to stay on the ITK side
* perhaps some of the interfaces should go into DCMTK, not this module's logic, but we don't have resources to work on factoring those out and migrating into DCMTK - we will have a better idea when we have legacy enhanced MR support ready in DCMTK

So at the moment, I am hesitant to put effort into factoring logic out. A relatively easy refactoring could be just restructuring the code to have a C++ library and a header, but that would operate on the level of DCMTK and ITK objects, so I am not sure it will serve the needs you have Csaba (I think we talked about this at the project week). But if that is all you need, let me know, that should be easy.

pieper:
The main reason for wrapping the current functionality in a logic class
would be to simplify/clarify/ruggedize the passing of metadata information
- the current info text files are imperfect for many reasons, but perhaps
mainly since they seem fragile.  Perhaps the best near-term use of effort
would be to move them to json just so the encoding is clearer.

cpinter:
The purpose of refactoring would be a more flexible and OOP-like usage of the converters. The CLIs may be convenient to implement and for batch processing, but especially the custom-made metadata file raises questions about its usability in general. It seems very hard-wired to me that the metadata file is incidentally created from within a python function in a not reusable way. Furthermore there is a lot of extra file IO happening in the background that is unnecessary if you work in Slicer. If there was a class that has functions that for example create an itkImage from DICOM SEG, create a metadata file for a DICOM UID in the database etc, then we wouldn't have to deal with temporary file pairs on local storage, and we could do these operations in memory, in a re-usable way.
I understand the QIICR point of view, but if we want to add features that can be used more generally than only in our own project, then we need that extra effort.

jcfr:
Agreed with @cpinter  points.

If functionality are currently in a CLI, first step for anyone willing to make it available would be to create a logic class used by that CLI and potentially re-usable from outside.

Having testing in place for that CLI would make that work easier and less error prone.

fedorov:
@cpinter @jcfr as I said before, I am fine with factoring out functionality into a C++ library that does not introduce any extra dependencies (Slicer logic, Slicer DICOM database, CTK, VTK, etc).

@cpinter what is a more reusable way to share metadata? The CSV file in a way parallels the "flat" "attributes" approach in SubjectHierarchy, Segmentations, and MRML nodes in general.

@pieper I agree JSON might be better, but what library support do we have for parsing JSON in C++? Are there any examples in Slicer? It is easy and fun to use JSON from python, but I am not writing a JSON parser in C++.

jcfr:
> for parsing JSON in C++

The most straightforward way would be to include JsonCpp either as a dedicated external project or by integrating the source like what is done for [CMake itself](https://github.com/Kitware/CMake/blob/master/Utilities/cmjsoncpp/README-CMake.txt). 

Note 1: JsonCpp is already CMake'ified. See https://github.com/open-source-parsers/jsoncpp

Note 2: @thewtex also did work creating the ParameterSerializer and integrate it to the SlicerExecutionModel to easily record CLI parameters into files and load them from files. See https://github.com/Slicer/SlicerExecutionModel/pull/20

cpinter:
The ParameterSerializer looks really promising. Can it serialize parameters other than what is described in the module XML? The main problem with the custom metadata file is that it's written and read by snippets of code in functions that do a lot of other stuff too, so it's not something that can be used from other places.

fedorov:
Let me think about all those pieces. This discussion is helpful

---

thewtex:
@cpinter Yes, in addition to the SlicerExecutionModel JSON serialization the ParameterSerializer is used for more detailed parameter specification.

> The main problem with the custom metadata file is that it's written and read by snippets of code in functions that do a lot of other stuff too, so it's not something that can be used from other places.

Yes, it would be nice to have SEM / ParameterSerializer output [JSON-LD](http://json-ld.org/) descriptions of their output (parameter names, descriptions, types, ...) in addition to a JSON output that may just be name / value pairs. I did not have time to implement this. This would facilitate use in other contexts.

pieper:
+1 for JSON-LD wherever possible.

fedorov:
Sorry for the long radio silence on this topic. Too many projects, and I don't have anyone else to assign to this at the moment. But I did not forget.

Here is how I am thinking to do:
* refactor the DICOM export/import functionality into a package that can be built both within and outside of Slicer (with dependencies on DCMTK, ITK, and SlicerExecutionModel)
* split the functionality into a library and general utility applications (SEG converter(s) being the first one). If you have specific suggestions what you would like API to look like - let me know!
* assign default values to as many parameters as possible to simplify usage
* consider supporting both CSV and JSON for communicating parameters
* consider creating a supporting tool that would spit out the CSV/JSON file with pre-populated default parameters to simplify command line usage

I did look into SlicerExecutionModel JSON support. This will not help here directly, since the number of parameters is not fixed and will depend on the number of labels in the input segmentation files (unless we want to remove that feature). As an aside, I also was not successful trying this out, since it is not enabled by default in Slicer or SlicerExecutionModel, and trying to enable it causes some build issues, at least on my mac - it may be it requires certain features of ITK enabled, which are not enabled in Slicer ITK build. Anyway, that was helpful since it pointed me to JsonCpp.

I will keep this thread updated as I make progress, an hopefully will have someone help me once another project reaches a milestone, but it is good to know that the functionality is already available and integrated in Segmentations module thanks to @Sunderlandkyl, for the immediate needs. Thanks for the contributions and feedback!

thewtex:
@fedorov That sounds like an awesome plan! +100

A note regarding JsonCpp -- when the SlicerExecutionModel JSON functionality was added, upstream JsonCpp did not have any CMake build support.  However, recently upstream finally add some support.  It is not perfect, but it does help.

A questions that comes up over and over again on the ITK mailing lists is how to add write out an image in DICOM format. This could have a huge impact if the core was written in classes in an ITK module.  There is now better documentation how to create a module:

  http://itk.org/ITKSoftwareGuide/html/Book1/ITKSoftwareGuide-Book1ch9.html#x50-1430009

and there are many examples of module development on GitHub, e.g.:

- https://github.com/InsightSoftwareConsortium/ITKIOTransformDCMTK
- https://github.com/InsightSoftwareConsortium/ITKMinimalPathExtraction
- https://github.com/InsightSoftwareConsortium/ITKAnisotropicDiffusionLBR
- https://github.com/InsightSoftwareConsortium/ITKCuberille
- https://github.com/InsightSoftwareConsortium/ITKTwoProjectionRegistration

fedorov:
@thewtex thanks a lot for these suggestions and pointers, especially ITKIOTransformDCMTK. This will be very helpful, and it would be another use case to exercise the API of this package/library. I completely agree about writing out a DICOM image. We are looking into this among many other things.

cpinter:
The plan makes good sense, thank you Andrey!

About the first point: How do you imagine this package? Would it be a separate git repo that is built as a project by the Reporting superbuild?

fedorov:
Csaba, Reporting is a legacy extension that is no longer needed. It used to provide a module that wrapped Editor and added few other things on top of it to support SEG export. It also provides SEG DICOM plugin. Now that SEG export is integrated natively (via Editor or Segmentations), that GUI is no longer needed. It also used to provide support for AIM, but that is not in the scope of our current project and we do not have resources (or use cases) to support it.

I think we should deprecate Reporting, whenever the new package is ready. I imagine that the package will have library and CLIs that can be used independently of Slicer, that would be discoverable by Slicer. The python plugin will also be there for now (we never made progress on what to do with DICOM plugins vs Slicer core, and how to facilitate their discovery when extension is not installed). The package should have a superbuild to detect when DCMTK/ITK/ExecutionModel are not available and to build those. I imagine other Slicer modules that provide Slicer UI should be separate and would just depend on this package.

cpinter:
So it will be part of Slicer core, but built separately similarly to SlicerDMRI?

The DICOM plugin discovery from not installed extensions seems to be a very difficult task, especially because there is no existing infrastructure that we can build on, so I don't really see that coming in the near future, but I guess @pieper and @jcfr can tell more about this.
I personally am not too worried about this, as most users seem to find SlicerRT for loading DICOM-RT. Not sure how they find it, probably by finding it in the extension category. On the other hand I think it's completely fine if SEG support is available by default in Slicer core, however more DICOM plugins mean more examine time, for example MultiVolumeImporter takes a long time to examine the files and I never use it.

FYI I have made good progress with the new Editor, maybe I can show it at a hangout soon.

fedorov:
> So it will be part of Slicer core, but built separately similarly to SlicerDMRI?

That is to be determined. In any case, that project is not ready. Let's discuss when the time is right.

> The DICOM plugin discovery from not installed extensions seems to be a very difficult task

What I meant is that CLIs can be used independently of Slicer, but if it is detected that the build is within the Slicer build, they would be packaged and installed.

> however more DICOM plugins mean more examine time, for example MultiVolumeImporter takes a long time to examine the files and I never use it.

If you never use it, why don't you disable it? I think the setting should stick.

cpinter:
Please let me know when the time is right.

I'm not familiar with the timeline of this project at all, but it may be possible that Segmentations is integrated before work on this topic is started. In that case the field is changed and we can implement a more comprehensive and permanent solution.

pieper:
I think it makes sense for DICOM plugins to be part of the core in almost
all cases.  But I also think that some of the processing code should come
from extensions and that we should provide a discovery mechanism to suggest
compatible modules/extensions when a user loads a particular data type (be
it from DICOM or some other means).  For CLIs we could automate some of
this by looking at the XML.  For extensions maybe we should provide a way
for the modules to be tagged.

cpinter:
I find it a really good idea to include all the DICOM plugins that can
1. Identify the modality using simple pydicom calls
2. Alert user if a supported modality is found but the module/library class is not present

fedorov:
> 1. Identify the modality using simple pydicom calls

@cpinter I wish looking at the modality tag was all that was needed by the plugins. In situations like multivolume, one has to look at various strategies to parse out the 4th dimension from the data, and that operation is not a simple lookup, that is why the plugin is slow (of course, I am sure it can be optimized).

pieper:
I agree with @fedorov - I wish it was simple to map dicom SOPClassUID to a
MRML node.  Sometimes it is and sometimes it is not.

It might also be nice if we could think of ways to do the examination step
automatically in the background, but that may not help when you goal is to
load new data files.

cpinter:
I did not mean direct mapping to different MRML types, but to determine from within the python DICOM plugin whether it can handle the examined file even without the more complex logic, and indicate its absence if that's the case.

Even if the examine step is very complex, giving the yes/no answer to the question "can I potentially load it" is simpler than fully filling the loadable. I guess you can tell better if it's simple enough in every case to be done from the plugin class itself, but it seems like a good idea to me.

pieper:
For some cases yes - for example the diffusion plugin has some logic based
on some tags that have to be present if dwi conversion is even remotely
possible, but you don't know for sure that it works until you run the
DWIConvert CLI.

Multivolume has similar complexities maybe it could follow a similar
pattern.  Basically offering to load with various 4th dimensions but then
providing feedback if the requested data is not found.

fedorov:
> giving the yes/no answer to the question "can I potentially load it" is simpler than fully filling the loadable.

Yes, I guess I can see how that could be done. But this would be a departure from our current conventions in DICOM plugins. Right now, if examine() results in a loadable, the user knows for sure that loadable can be loaded (it has a confidence, but it is not the confidence of it being loaded, but the confidence of it being the correct interpetation of a DICOM data). If I understand correctly, you suggest to introduce a different kind of confidence of how likely the initial guess would lead to a valid loadable.

cpinter:
> you suggest to introduce a different kind of confidence of how likely the initial guess would lead to a valid loadable

Steve's idea about keeping the DICOM plugins in the core and the discovery mechanism seemed worthwhile to brainstorm about, but I don't know enough about all DICOM modalities to suggest a good solution. As I wrote earlier this seems like a long-term project.

fedorov:
I completely agree this is something we should keep brainstorming!

So as a continuation on this idea, maybe we should

1. have an external project (like MultiVolumeExplorer or SlicerDMRI) to host all DICOM plugins
2. update all plugins so that they have checks whether the module handling the logic implemented outside the python code installed
3. if the logic is not installed, give a guess and notify the users which extension needs to be installed to proceed with the guess and try to interpret the series better

I think this should be reasonably easy. The notification of the missing extension would need to be added to the DICOM module, but the rest is minimal changes to the plugins.

cpinter:
Looks like a good plan to me! I'll add a Mantis issue for this if there are no objections, and we can get back to this when it becomes priority for someone.

About the SEG import library, Andrey will see if and when they can work on outsourcing the code and deprecate Reporting. As the SEG import feature works fine thanks to Kyle, there is no big rush on this matter. In the meantime I will work on the prerequisites to integrate Segmentations to the Slicer core, and whoever is there first, will let the others know.

Does this sound okay?

pieper:
Works for me

cpinter:
I created issue http://www.na-mic.org/Bug/view.php?id=4146 to track the DICOM plugin discovery feature.

fedorov:
Thanks Csaba. I will send email to slicer-devel later to make sure people have a chance to comment on this.

jcfr:
> I will work on the prerequisites to integrate Segmentations to the Slicer core, and whoever is there 
> first, will let the others know.

> Does this sound okay?

:+1:

fedorov:
Hi guys, few quick updates on this topic:

* DICOM conversion functionality is now available in a standalone dcmqi library that is a superbuild of ITK, DCMTK and SlicerExecutionModel, no other dependencies: https://github.com/qiicr/dcmqi
* continuous integration for all 3 platforms is fully functional
* jsoncpp is integrated as part of the library
* @pieper is working on improving the Slicer plugin infrastructure along the lines discussed in http://www.na-mic.org/Bug/view.php?id=4146
* @che85 is engaging with this project, so we should expect a lot quicker progress
* @jcfr is getting on-board with dcmqi and will be helping with cmake, superbuild, installation, packaging and similar issues
* QIICR deadline is to have a refactored code and new Slicer extension modules leveraging the updated functionality ready in for the demo at RSNA in Nov 2016

jcfr:
Nice summary

> jsoncpp is integrated as part of the library

Since it is also needed by the ParameterSerializer in SlicerExecutionModel, I suggest we add JsonCpp into Slicer core. 

> Re: http://www.na-mic.org/Bug/view.php?id=4146

What about:
1) updating the Slicer module interface so that Slicer module can declare metadata
2) updating the extension description file so that it summarizes the metadata of each bundled modules
3) update slicer extension manager so that it can download (or be packaged with) the list of all extension description along with metadata ?

pieper:
I would really like to see us move to json for things in the core, so +1 to jsoncpp.

Regarding the issue 4146 discussion: yes, Andrey and I are scheduled to meet on Thursday at 1pm to discuss how general to make this feature (should it just be DICOM or should other attributes of extensions be describable).  If you could join @jcfr that would help, otherwise we can discuss during the developer hangout tomorrow.

cpinter:
Currently the DICOM plugins have both import and export capabilities. Are you only interested in providing this exploration feature only for import or both?

pieper:
Right now we are looking at import only.  I suspect that the extensions
will provide several features all in a single package, so that if you
install, e.g. SlicerRT gives you importers but you also get an exporters.
Same would be true for a SEG, at least as we have things now.

That's why this could be such an open-ended feature to add.  Basically if
the user says "here's my data, how can I load it?" it sort of naturally
implies a series of questions like "what modules are compatible with my
data?" and "how will I be able to save my data?"

I think metadata associated with the extensions can be a big help with
this, but it's not clear how far we should go.  I'd like to keep the scope
very narrow (just dicom SOPClassUIDs or similar) unless we see a way to add
a lot of value to the user.

cpinter:
This makes sense. Import and export usually come from the same place. A simple infrastructure with SOPClassUIDs or tags would suffice I think.

Not sure what to do about the case when there is an importer for a peculiar type of a known modality. An example is the variable slice thickness CT import in Plastimatch that we will probably add to SlicerRT some time.

Also it could happen that two extensions support a modality, but one is more comprehensive. My example for this is SRO support in SlicerRT that is only partially done (rigid only for now). Should we give confidence values for each SOPClassUID meaning how "well" the extension supports the modality?

cpinter:
(I can join tomorrow's hangout if this will be a topic and skip the long discussions here for now)

fedorov:
I am traveling tomorrow, but will discuss with Steve and later with JC.

jcfr:
Developer meeting page for tomorrow has been created. I added this as an optional topic for preliminary discussion.

See http://wiki.slicer.org/slicerWiki/index.php/Developer_Meetings/20160425

pieper:
I added a proposal to the mantis issue based on my discussion with Andrey.  I want to get something useful implemented but not try to tackle the whole huge issue.  We can always do something more in the future.

http://www.na-mic.org/Bug/view.php?id=4146#c13865

PR #72, opened 2016-05-31
STYLE: Explicitly import "vtkTeem" to use "vtkImageLabelCombine"

jcfr:


jcfr:
Integrated this PR means that Slicer version including this PR will be required: https://github.com/Slicer/Slicer/pull/507

fedorov:
no longer needed, this is concerned with the old label infrastructure no longer used

PR #73, opened 2016-05-31
BUG: Simplify import of Logic/MRML/DisplayableManager python modules

jcfr:
Reference wrapped module directly from "slicer" module to adapt to
Slicer core change introduced in r25128.


fedorov:
this is concerned with the deprecated code now removed ...

Issue #74, opened 2016-09-13
Design/requirements for the initial version of the new Reporting module


fedorov:
There should be different modes of operation for annotation vs viewing (from existing SR data).

Annotation mode:
- [ ] ~~Save intermediate results as the user does annotation for fault-tolerance (not DICOM representation in the user-visible database interface)~~
- [ ] When user explicitly confirms completion or exists the module, export SEG and SR - when user finalizes
- [ ] (note for the future) maintain predecessor reference (DICOM thing) - for SR only, perhaps can be used to differentiate SEGs as well
- [ ] If predecessor concept is used, only the final version of the report is exposed to the user in the interface
- [x] Import DICOM image series
- [x] On entering the module - prompt the user what datasets in the scene can be annotated (only those that were imported from DICOM)
- [x] Use Segmentations Editor for annotation
- [x] Only volumetric/pixel labeling annotations
- [x] Provide interface for selecting terminology of the structure being annotated (Csaba is working on this)
- [x] Calculate measurements when interaction is done - just the basic measurements (as in Label Statics)
- [x] use Table node for storing the measurement results, displaying the table and organizing the measurements in the SubjectHierarchy module
- [x] SR complete/partial - verified/unverified - initially, set it to complete+verified when user is done
- [x] when user clicks on the structure in the list, re-center the view 
- [x] take the best features as makes sense from the mpReview module (center on click, solid vs outline shortcuts, window/level)

fedorov:
let's do prototype in this one? http://pencil.evolus.vn/Features.html

fedorov:
initial iteration, done in Pencil

![reporting_v2_sketch](https://cloud.githubusercontent.com/assets/313942/18537015/41be8b14-7ad0-11e6-8d1f-20cfc7075f4d.png)

Source file and screenshot are also available https://github.com/fedorov/Reporting/tree/v2-design/Doc/design

@che85 can you join the QIICR hangout tomorrow to have some brainstorming about this? If you have any suggestions/iterations before that, they are definitely welcomed, but I prefer to have an early prototype and early discussion then waiting for another 2 weeks.

che85:
Looks good to me. Anyway it was better, that you create the first draft because you had a more concrete idea of all this than me. I will be joining the QIICR call.

fedorov:
I agree - that's what I thought too!

fedorov:
Notes from the discussion at the QIICR hangout Sept 16, 2016:

[...]

PR #75, opened 2016-10-04
New reporting module prototype

che85:
for the reference #74

depends on https://github.com/SlicerProstate/SlicerProstate/pull/38
depends on https://github.com/QIICR/dcmqi/pull/107

@fedorov work in progress

### DIFF
--- a/Py/Reporting.py
+++ b/Py/Reporting.py
@@ -1,6 +1,7 @@
 import getpass
 import json
 import logging, os
+import datetime
 
 from slicer.ScriptedLoadableModule import *
 import vtkSegmentationCorePython as vtkCoreSeg
@@ -292,16 +293,13 @@ class ReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidget):
       slicer.app.applicationLogic().PropagateTableSelection()
 
   def onSaveReportButtonClicked(self):
-    dcmSegmentation = self.createSEG() # TODO: save to DICOM database and retrieve UID
-    # TODO: Save Structured Report to DICOMDatabase
-    # self.createDICOMSR(dcmSegmentation)
-    print "on save report button clicked"
+    dcmSegmentation = self.createSEG()
+    self.createDICOMSR(dcmSegmentation)
 
   def onCompleteReportButtonClicked(self):
     print "on complete report button clicked"
 
   def createSEG(self):
-    # TODO: Create Json
     data = dict()
     try:
       data.update(self._getSeriesAttributes())
@@ -314,41 +312,64 @@ class ReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidget):
     logging.debug("DICOM SEG Metadata output:")
     logging.debug(data)
 
-    segmentationVolume = slicer.vtkMRMLScalarVolumeNode()
-
-    segmentationVolume.SetName("Segmentation")
-
     tempDir = slicer.util.tempDirectory()
 
     labelNode = self.segmentEditorWidget.logic.labelMapFromSegmentationNode(self.segmentEditorWidget.segmentationNode)
     slicer.mrmlScene.AddNode(labelNode)
     slicer.util.saveNode(labelNode, os.path.join(tempDir, "labelmap.nrrd"))
 
-    outputSegmentatonPath = os.path.join(tempDir, "segmentation.dcm")
+    currentDateTime = datetime.date.today().strftime("%Y%m%d")
+
+    metafilePath = self.saveJSON(data, os.path.join(tempDir, "meta.json"))
+    outputSegmentatonPath = os.path.join(tempDir, "seg_{}.dcm".format(currentDateTime))
 
     params = {"dicomImageFiles": ', '.join(self.getDICOMFileList(self.segmentEditorWidget.masterVolumeNode,
                                                                  absolutePaths=True)).replace(', ', ","),
               "segImageFiles": labelNode.GetStorageNode().GetFileName(),
-              "metaDataFileName": self.saveJSON(data, os.path.join(tempDir, "meta.json")),
+              "metaDataFileName": metafilePath,
               "outputSEGFileName": outputSegmentatonPath}
 
     logging.debug(params)
-    slicer.cli.run(slicer.modules.itkimage2segimage, None, params, wait_for_completion=True)
+
+    cliNode = None
+    cliNode = slicer.cli.run(slicer.modules.itkimage2segimage, cliNode, params, wait_for_completion=True)
+    waitCount = 0
+    while cliNode.IsBusy() and waitCount < 20:
+      slicer.util.delayDisplay("Running SEG Encoding... %d" % waitCount, 1000)
+      waitCount += 1
+
+    if cliNode.GetStatusString() != 'Completed':
+      raise Exception("encodeSEG CLI did not complete cleanly")
 
     if not os.path.exists(outputSegmentatonPath):
       raise RuntimeError("DICOM Segmentation was not created. Check Error Log for further information.")
-    indexer = ctk.ctkDICOMIndexer()
-    indexer.addFile(slicer.dicomDatabase, outputSegmentatonPath)
-    return segmentationVolume
+    slicer.dicomDatabase.insert(outputSegmentatonPath)
+    return slicer.dicomDatabase
 
-  def createDICOMSR(self, referencedSegmentation):  # TODO: we might have several segmentations
-    data = dict()
-    data.update(self._getSeriesAttributes())
-    # compositeContextDataDir, data["compositeContext"] = self.getDICOMFileList(referencedSegmentation)
+  def createDICOMSR(self, referencedSegmentation):
+    return
+
+    data = self._getSeriesAttributes()
+    compositeContextDataDir, data["compositeContext"] = self.getDICOMFileList(referencedSegmentation)
     imageLibraryDataDir, data["imageLibrary"] = self.getDICOMFileList(self.segmentEditorWidget.masterVolumeNode)
 
     print "DICOM SR Metadata output:"
-    print json.dumps(data, indent=2, separators=(',', ': '))
+    logging.debug(data)
+
+    json.dumps(data, indent=2, separators=(',', ': '))
+
+    logging.debug(params)
+    cliNode = None
+    cliNode = slicer.cli.run(slicer.modules.tid1500writer, cliNode, params, wait_for_completion=True)
+    waitCount = 0
+    while cliNode.IsBusy() and waitCount < 20:
+      slicer.util.delayDisplay("Running SR Encoding... %d" % waitCount, 1000)
+      waitCount += 1
+
+    if cliNode.GetStatusString() != 'Completed':
+      raise Exception("encodeSEG CLI did not complete cleanly")
+    # TODO: Save Structured Report to DICOMDatabase
+
 
   def _getSeriesAttributes(self):
     return {"SeriesDescription": "Segmentation",
@@ -587,7 +608,10 @@ class ReportingSegmentEditorLogic(ScriptedLoadableModuleLogic):
   def labelMapFromSegmentationNode(self, segNode):
     labelNode = slicer.vtkMRMLLabelMapVolumeNode()
     slicer.mrmlScene.AddNode(labelNode)
-    if not self.segmentationsLogic.ExportAllSegmentsToLabelmapNode(segNode, labelNode):
+
+    mergedImageData = vtkCoreSeg.vtkOrientedImageData()
+    segNode.GenerateMergedLabelmapForAllSegments(mergedImageData, 0)
+    if not self.segmentationsLogic.CreateLabelmapVolumeFromOrientedImageData(mergedImageData, labelNode):
       slicer.mrmlScene.RemoveNode(labelNode)
       return None
     return labelNode
--- a/Py/Reporting.py
+++ b/Py/Reporting.py
@@ -1,6 +1,4 @@
 import getpass
-import SimpleITK as sitk
-import sitkUtils
 import json
 
 from slicer.ScriptedLoadableModule import *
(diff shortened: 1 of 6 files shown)


fedorov:
Is everything in the repo?

image

fedorov:
The only plugin in Py/DICOMSRPlugin.py is DICOMSRHTMLPlugin

che85:
Interesting. Let me check

che85:
Did you build it or did you just add the py directory into your Slicer module settings?

fedorov:
I just added the py directory. I don't have a recent Slicer build, but I can make one if that is needed. I don't see however how it could solve this particular issue.

Sorry I was going to get started with this yesterday, but only getting to it now ...

che85:
Same here. I cannot test it, as long as my Slicer build is running.

fedorov:
ok, let me start a build too

che85:
For me it works, when I build it

fedorov:
do you know why? where is the class DICOMSRPlugin defined?

che85:
I think it has something to do with the directory structure ifself. I will take a look.

che85:
I have no DICOMSRPlugin.py in that Py directory see https://github.com/che85/Reporting/tree/newReporting/Py

fedorov:
oh, of course - sorry, this is totally my fault ...

image

che85:
But still the other Error appears also on my machine, but it doesn't when I build Reporting

che85:
When I delete file qSlicerReportingModuleWidget.py, then it works. Let me create a new commit

che85:
But even with the error you should still be able to select it. I think we need to clean up the directories in the future...

fedorov:
Small issues to be fixed for the initial merge:

- [ ] Everything should be disabled until measurement report is created, measurement report should allow creating new report
- [ ] While in the W/L mode, selection of the tools from Editor do not work, because of a conflict with the W/L mode I guess
- [ ] Need to maintain reference to the master volume node from the segmentations
- [ ] In my tests, automatic calculation of the measurements was not triggered

che85:
@fedorov What should be the behavior when W/L mode is activated? Should the Editor be disabled? Should W/L not be enabled when Editor is active?

fedorov:
I think it makes sense W/L to behave as an Editor effect, when Editor is used, eg, when an Editor effect is turned on, it automatically disables W/L, and when W/L is triggered, it disables the effect.

However, for this initial use case, where we do not need Fg layer, it makes sense to not add this W/L button at all, to keep things simple, and use standard out of the box W/L function.

che85:
Ok I am going to remove it for now

che85:
@fedorov You should be able to test it again. 

@cpinter I noticed, that while creating a segmentation, a reference to the master volume is created somehow. When creating another segmentation node and then switching back to the first created one, the original master volume is selected automatically. I am wondering where 
you save a reference to the master volume specifically for the segmentation node.
Is it possible to access this via python?

cpinter:
I cannot mention Andras (@lassoan) because probably he's not a member in this space, but he did a major revamp of the labelmap handling in Segment Editor afer I finished the first round of implementation. It included handling of master volumes, so he will be able to answer immediately, while I'd need to read code and figure it out just as you would need to.

Please invite Andras here so he can take part of the conversation about those components that he created or modified himself.

cpinter:
Maybe the mention worked after all? It was not suggested in the popup list.

lassoan:
I would suggest to use the "NodeI info" module in DebuggingTools extension to explore node contents and references (you can traverse through node references, open watch windows, force node modifications, inject variables to the Python console, etc).

To answer your specific question: it's the "referenceImageGeometryRef" (you can access this name by calling vtkMRMLSegmentationNode::GetReferenceImageGeometryReferenceRole()) node reference that indicates which volume's geometry to use.

lassoan:
I've subscribed to this ticket, so I get notified about posts.

fedorov:
Csaba, mentions always work. I do not know how suggestions are selected, they are not always intuitive. In either case, it makes sense to add Andras to the team, I've just invited him.

fedorov:
> To answer your specific question: it's the "referenceImageGeometryRef" 

For the legacy Editor, we used attribute "AssociatedNodeID". 

I also found just now that when a segmentation is imported from DICOM SEG, it does not seem to keep any attributes pointing to the referenced volume (tested with the Oct 8 nightly), will need to fix this.

cpinter:
Yes, it will be needed to fix for sure. Thanks for adding the ticket!

The node reference will be better than the AssociatedNodeID attribute, as there can be multiple nodes referenced using the same reference. Would it be needed to have a current master node ref?

che85:
@lassoan Does it also keep track of different master volume that I used to create segmentations or is it always only the very first one?

lassoan:
Attributes are not suitable for keeping reference to nodes, because when you import a scene then node IDs may change. Node references keep track of these node ID changes, while not attributes are simple strings, which may end up pointing to the wrong nodes.

To keep track of DICOM associations, we need to use DICOM UIDs, as the referenced image may be loaded later or not at all, while node references are only valid between nodes in the scene.

lassoan:
It's the segment editor's responsibility to modify the geometry of the segmentation. Currently, it sets the geometry once (with the first volume set as master) and it is kept as is forever. It would be useful to expose this to the user and allow the user to change it.

The matter of storing references to each volume that was used for segmentation is a completely different story. Probably we should store a list of DICOM UIDs of all the volumes that were displayed when changes were made to the segmentation. We could use segment tags or segmentation node attributes for that.

fedorov:
> Attributes are not suitable for keeping reference to nodes, because when you import a scene then node IDs may change. 

Our general principle is to never rely on Slicer scene when reloading the data. In our workflow, keeping node ID is acceptable, since we rely on it only during the given session. Keeping a reference to the SeriesInstanceUID is also fine, as far as I am concerned, and I don't know if there is any other application that uses this feature anyway (other than SubjectHierarchy perhaps).

About handling multiple referenced volumes - as I mentioned before, let's leave this out of the use case/workflow. We can discuss how it can be incorporated at the project week. For now let's always assume we have only one master.

che85:
@fedorov I am not sure what values to take for Finding and FindingSite https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-schema.json#L49-L54 for creating a DICOM SR. I assume, that I would take some of the terminology used for creating segments, but I am not sure and I wanted to ask before I code that. 

I also don't know what values to take for MR. This shows values for CT https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-ct-liver-example.json#L102-L111 but I couldn't find any codes for MRI (only Multiparametric MRI http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_126020)


I noticed, that you were using the wrong code value here: https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-ct-liver-example.json#L85-L87
which is actually this one : http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_122713

I assume, this is the correct one: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_112031

che85:
@fedorov I know that you are talking about Finding and FindingSite in https://peerj.com/articles/2057/#p-37, but for me is not clear, where we get the anatomical location get from, when for instance no anatomical context is available for a segment. To me is unclear, how to get this location. 

For Finding, I would assume to take the SegmentedTypeProperty...

fedorov:
> @fedorov I am not sure what values to take for Finding and FindingSite https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-schema.json#L49-L54 for creating a DICOM SR. I assume, that I would take some of the terminology used for creating segments, but I am not sure and I wanted to ask before I code that.
> 

I think it makes sense to assign Finding to have the value of SegmentedPropertyType.

FindingSite should not be initialized in the instances when AnatomicLocation in SEG is not initialzied.

> I also don't know what values to take for MR. This shows values for CT https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-ct-liver-example.json#L102-L111 but I couldn't find any codes for MRI (only Multiparametric MRI http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_126020)
> 

For MRI, the quantity should have code `(DCM, 110852, MR signal intensity)`, and units are `(1, UCUM, "no units")`.

In the future, in other applications, we can be more specific when we know more about the context, but for now "MR signal intensity" is good.

> I noticed, that you were using the wrong code value here: https://github.com/QIICR/dcmqi/blob/master/doc/sr-tid1500-ct-liver-example.json#L85-L87
> which is actually this one : http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_122713
> 
> I assume, this is the correct one: http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_D.html#DCM_112031

You are right. I took the code without checking from an email communication with David, and it does look like David made a mistake there!

che85:
@fedorov I created successfully a DICOM SR for the first time. dciodvfy doesn't display any errors...

I hardcoded some values which I don't know where to populate from:
For SEG:
- ClinicalTrialSeriesID
- ClinicalTrialTimePointID
- ClinicalTrialCoordinatingCenterName

For SR:
- activitySession
- timePoint

"ObserverType" will always be "PERSON" I guess (at least for now)

fedorov:
> I hardcoded some values which I don't know where to populate from:
> For SEG:
> 

Great, thanks!

> ClinicalTrialSeriesID
> ClinicalTrialTimePointID

I think these are not important for now, we can always assign "1"

> ClinicalTrialCoordinatingCenterName

You can put "QIICR" as default

> For SR:
> 
> activitySession
> timePoint

These can also be just 1

> "ObserverType" will always be "PERSON" I guess (at least for now)

Yes

lassoan:
Do you use Slicer's DICOM export infrastructure? That has nice way of specifying required tags that the users can edit as they see fit (GUI is generated automatically).

che85:
You mean from here? https://github.com/QIICR/Reporting/blob/master/Py/DICOMSegmentationPlugin.py

Right now, I just coded everything in order to be able to have a first prototype. I plan to do refactoring/outsourcing afterwards.

fedorov:
I have reservations about that interface. I will explain later. Christian let's discuss before you do anything in that direction.

lassoan:
The interface can be improved. If by investing a bit more time e can get more generic, reusable, flexible solution then I would prefer that. Could we have a quick discussion about design options?

fedorov:
Andras, the issue is not so much about the details of the interface, but about the conceptual approach about how you do DICOM export.

I understand your desire to have generic reusable solutions, but QIICR aims to develop radiology review workflow and clinicians (or someone close anyway) as the target users. What Slicer experience shows over and over is that generic components create tremendous confusion to the non-technical users (think Data save/load, MRML scene, Editor, Volumes). If you look at the commercial solutions, they develop highly customized workflows for specific tasks. I can see how Editor, Subject Hierarchy etc can be helpful for a user like you and me, but not for a user like Fiona or Clare.

I want the operator to know only what is absolutely necessary about DICOM. I absolutely do not want the user to edit DICOM attributes directly, or to even consider that, not in the workflow we want to implement. This is why I want to populate as much as possible automatically, and where automatic is not possible, have GUI elements directly in the workflow to expose those controls.

lassoan:
Yes, I agree that we need to provide very simple user interfaces and workflow for clinical use. 

Implementation is always a mix between reusing customizable Slicer core components and developing new application-specific components. So, we need to find a good balance between
A. spending time with improving Slicer core components so that it can fulfill application-specific requirements, and
B. spending time with implementing application-specific code.

A. is waste of resources in the short term: it takes maybe 2-3x more time to develop something that is robust and flexible enough for reuse. B. is a waste of resources in the long term: needs of different groups are often very similar, so different groups overall spend 5-10x more time with reimplementing, testing, and maintaining the same thing multiple times. I'm advocating to try to aim for developing/improving core components and keeping application-specific layer very thin, because in a few-year timeframe this provides more return on the investment.

I think we are doing OK overall - reusing segmentations infrastructure, tables, adding terminology to the core, etc. are all very good in the long term.

fedorov:
Yes, I agree that reuse is good, and we do reuse and factor out things as suitable, and when we have resources. However, there also limits to reuse. DICOM Export is one instance where I am not convinced reuse is good for supporting end user we have in mind.

Along the lines you discussed, we should review the generic components Christian had to improve for the prostate applications, perhaps at the project week. There are many basic things that are not done right (very user unfriendly) in the core and have to be fixed in the extensions. Christian did a lot of work in that direction, but we do not have the resources to move all those improvements into the core, so it would be good to discuss and prioritize, and see who can contribute resources.

che85:
@fedorov I noticed, that a DICOM SR can have several Observers. For now the JSON schema only supports one observer. Should I correct that or do we assume that there is only one observer (for now)?

fedorov:
Why did you conclude that? According to the TID 1002, VM=1 http://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_A.html#sect_TID_1002

image

che85:
Interesting. The code from the dcmtk says the following:

image

Or it's just about the number of verifications. But still this is important, right? 

But as far as I remember, we decided that a DICOM SR can reference other (previous) DICOM SRs so then I might be wrong.

fedorov:
Ah, I see - yes, I can see how the number of verifying observers can be more than one. But I have not looked into how exactly this would be represented in the document. There must be a separate template for verifying observers in addition to the author. Can you trace it up the tree? But I think for now we don't need to worry about this. We didn't do it for the Iowa dataset.

Py/Reporting.py
```diff
@@ -0,0 +1,885 @@
...
+
+    if not os.path.exists(outputSegmentationPath):
+      raise RuntimeError("DICOM Segmentation was not created. Check Error Log for further information.")
+    indexer = ctk.ctkDICOMIndexer()
+    indexer.addFile(slicer.dicomDatabase, outputSegmentationPath)
+
+    logging.debug("Saved DICOM Segmentation to {}".format(outputSegmentationPath))
+    return outputSegmentationPath
+
+  def createDICOMSR(self, referencedSegmentation):
```
fedorov: should names be consistent? Right now it is `createSEG` vs `createDICOMSR`

che85:
depends on https://github.com/Slicer/Slicer/pull/622

fedorov:
@che85 I think it is time to merge this into master. What do you think? This way it will also be easier to clean up the directory from the old stuff, and we could possibly contribute together more effectively.

Also, we discussed this module with the QIICR team, and David and Steve were saying the name may be confusing. How about renaming to "Quantitative Reporting"? (we can do this after merge)

che85:
Sure feel free to merge.... I will be working on the DICOMSegmentationPlugin to adapt to dcmqi and after that I will hopefully get everything running with reading a DICOM SR into "Quantitative Reporting" and displaying it

PR #88, opened 2016-11-17
STYLE: remove unused files

fedorov:
Also, added new DICOM plugins to Py/CMakeLists.txt

@che85 if you are working tonight, please take a look in the next 15 minutes. I wanted to merge before 10pm to make it into the nightly. I will fix tomorrow if I messed something up!


che85:
Wow. That's a lot. I will take a quick look.

fedorov:
It was all quite straightforward. I left renaming for later (and, as usual, documentation etc ....).

Perhaps most important is that the new DICOM plugins were not included in CMakeLists.txt, so they would not be in the nightly.

fedorov:
when done - please merge!

che85:
Ok LGTM. You should better know which files we don't need anymore. Basically most or all of it seems to be covered by dcmqi now.

fedorov:
Thanks!

PR #97, opened 2016-11-21
ENH: Terminology information from DICOM SEG is now added to the segme…

cpinter:
…nts, so segments table views can now handle that

### DIFF
--- a/Py/DICOMSegmentationPlugin.py
+++ b/Py/DICOMSegmentationPlugin.py
@@ -4,6 +4,7 @@ import string
 import vtk, qt, ctk, slicer
 from DICOMLib import DICOMPlugin
 from DICOMLib import DICOMLoadable
+import logging
 
 #
 # This is the plugin to handle translation of DICOM SEG objects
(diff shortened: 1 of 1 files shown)


fedorov:
@cpinter great, thanks!

What is the Slicer dependency? You mentioned you did a lot of work in the main application, but I don't recall seeing pull requests.

fedorov:
oh, ok - looks like you committed directly - can you confirm I need to build this hash https://github.com/Slicer/Slicer/commit/9c318571b60ddb67944409e73fbe2f352c9076be?

I might just go ahead and merge it, and test in tomorrow's nightly.

cpinter:
Yes, that's the commit in Slicer core

fedorov:
Looks pretty good, thanks!

One thing that is missing is that anatomic context is not populated from json, but I think this can be fixed later. So I am going to merge and see how things look in the binary tomorrow.

Also, looking at the output, I see that meta.json actually encodes all segments. There is no separate metadata json for each output segment NRRD as I mistakenly thought. Sorry for confusing you.

This dataset might be handy for testing: https://fedorov.gitbooks.io/rsna2016-qirr-dicom4qi/content/instructions/seg.html#test-dataset-3

PR #102, opened 2016-11-23
Enabling read only mode for reports read from SlicerDICOMDatabase

### LINKED ISSUE #90: Enable readonly mode for measurement reports
After loading a DICOM SR into Slicer it shall be available for selection in Reporting. 

This is also applicable to completed measurement reports.

Once selected, it should be read only:

- [x] Disable SegmentEditor
  - [x] No adding/removing of segments (@che85)
  - [x] Selection of segments should be possible, but not editable (@cpinter)
    - [x] No changes of segment name or terminology 
- [x] Reporting Buttons shall be disabled (@che85)

@fedorov Feel free to add items or change this list
@cpinter Is it possible to disable double click event listeners for segment name and terminology widget? Or at least forbid the change somehow?

### LINKED ISSUE #103: Cleanup code and remove temporary files after DICOM SEG and SR creation

### LINKED ISSUE #106: Support clicks in the measurements table to select segments
Would be nice to be able to click on the rows of the measurements table and have the same effect as selecting segments from the segment list.

### LINKED ISSUE #107: Improve user access to long lists of segments in read-only mode
With the segment list shown in read-only mode, it is impossible to see (or click!) segments that are not showing up in the very small area allocated to the list.

@cpinter is it possible to make individual segments not editable, but make it possible to scroll down the list?

@che85 @cpinter we should not display the Segment Editor controls that are irrelevant in the read-only mode (see screenshot with the red square around the stuff that should not be shown)

image

### LINKED ISSUE #110: Add cross-hair for the selected segment?
In the situations where multiple segments are visible on the same slice, it might be helpful to unambiguously highlight the region. 

It looks like the easiest way to do this is to enable slice intersections. May not need any changes, or add a shortcut button as in SliceTracker

image

che85:
- some minor refactoring by extracting a new method and removing unused variables

fixes #90
fixes #103
fixes #106
fixes #107
fixes #110

### DIFF
--- a/Py/Reporting.py
+++ b/Py/Reporting.py
@@ -238,6 +238,7 @@ class ReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidget):
 
   @postCall(refreshUIElementsAvailability)
   def onImageVolumeSelected(self, node):
+    self.seriesNumber = None
     self.initializeWatchBox(node)
 
   @postCall(refreshUIElementsAvailability)
@@ -447,10 +448,10 @@ class ReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidget):
 
   def _getSeriesAttributes(self):
     attributes = {"SeriesDescription": "Segmentation"}
-    seriesNumber = ModuleLogicMixin.getDICOMValue(self.watchBox.sourceFile, DICOMTAGS.SERIES_NUMBER)
-    # TODO: keep counter for the exported segmentations to avoid duplicate
-    attributes["SeriesNumber"] = "100" if seriesNumber in [None,''] else str(int(seriesNumber)+100)
-
+    if not self.seriesNumber:
+      self.seriesNumber = ModuleLogicMixin.getDICOMValue(self.watchBox.sourceFile, DICOMTAGS.SERIES_NUMBER)
+    self.seriesNumber = "100" if self.seriesNumber in [None,''] else str(int(self.seriesNumber)+100)
+    attributes["SeriesNumber"] = self.seriesNumber
     attributes["InstanceNumber"] = "1"
     return attributes
 
--- a/Py/Reporting.py
+++ b/Py/Reporting.py
@@ -168,6 +168,7 @@ class ReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidget):
     self.fourUpSliceLayoutButton = FourUpLayoutButton()
     self.fourUpSliceTableViewLayoutButton = FourUpTableViewLayoutButton()
     self.crosshairButton = CrosshairButton()
+    self.crosshairButton.setSliceIntersectionEnabled(True)
 
     hbox = self.createHLayout([self.redSliceLayoutButton, self.fourUpSliceLayoutButton,
                                self.fourUpSliceTableViewLayoutButton, self.crosshairButton])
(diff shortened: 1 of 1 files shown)


Py/Reporting.py
```diff
@@ -450,11 +449,7 @@ def _getSeriesAttributes(self):
     attributes = {"SeriesDescription": "Segmentation"}
     seriesNumber = ModuleLogicMixin.getDICOMValue(self.watchBox.sourceFile, DICOMTAGS.SERIES_NUMBER)
     # TODO: keep counter for the exported segmentations to avoid duplicate
-    #  series numbers
-    if seriesNumber in [None,'']:
-      attributes["SeriesNumber"] = "100"
-    else:
-      attributes["SeriesNumber"] = str(int(seriesNumber)+100)
+    attributes["SeriesNumber"] = "100" if seriesNumber in [None,''] else str(int(seriesNumber)+100)
```
fedorov: I know this is my code! :)

fedorov:
Thanks! I will test and report back.

fedorov:
@che85 SR series descriptions says "Segmentation"

che85:
What would you want it to be ? "Structured Report"

fedorov:
Let's call it "Measurements Report"

fedorov:
As discussed with @che85 in an off-github communication, reload of SR works, but jump to second segment on entry selection in segment list doesn't.

che85:
For computation of the segment's centroid, we need a binary labelmap, which I expected to get with calling 

```
self.segmentationsLogic.GetSegmentBinaryLabelmapRepresentation(segmentationNode, segmentID, imageData)
```

For the first segment, it works fine and the pixel value of the segment is 1.
The second segment's pixel value is 2 ....

@cpinter, @lassoan Am I missing something?

fedorov:
@che85 so the issue is that you expect 1 and instead it assigns the segment ID?

Maybe as a practical workaround you can just threshold the content of the label image for the segment, and assign 1 to the values above 0, and then compute the centroid on the result?

che85:
Yeah I was thinking about that too. Maybe I am wrong, but a my understanding of a binary label map would be 0 and 1. Otherwise the method name is misleading

fedorov:
Not sure, it depends on the definition. Binary may mean 0 or non-zero. In the DICOM SEG language, BINARY is opposite to FRACTIONAL, where fractional means pixel values encode probabilities or fractional occupancy of pixels. So I think 1 or segment ID - the difference is no big deal.

cpinter:
For binary labelmaps the point is that the foreground is non-zero. It might have many reasons why it's 2 and not one, but in general we try to restrict foreground to value of 1. How did you acquire this segmentation?

che85:
Ok I will commit soon. I fixed it.

fedorov:
Follow-up issues identified during review are #104 #105 #106 #107

fedorov:
@cpinter you are right - it is the reading issue! Closes #109

lassoan:
Maybe it's documented well (or at all) but in binary labelmap representations voxel that have <=0 value belong to background, >0 belong to the object. If you need a specific value (e.g., 1) then you have to apply thresholding.

che85:
@lassoan I already committed code for using a threshold. Thanks for the followup

fedorov:
@che85 that's the problem with cross-hair - it basically doesn't have any effect unless you synchronize slices. That is why I like the slice intersection mode better!

fedorov:
Looking good @che85, although I don't think it fixes #110. But good to merge anyway!

PR #108, opened 2016-11-23
ENH: TID1500 can be either Enhanced or Comprehensive SR

fedorov:


fedorov:
@che85 I am merging this since I thnk this is very minor and should not affect anything else

PR #112, opened 2016-11-23
Setting table height and synchronizing selection in both segment and measurments table

che85:


fedorov:
Very nice, and much better use of space! 👍

Issue #104, opened 2016-11-23
Newly added segments should have terminology populated


fedorov:
It is highly non-intuitive that newly added segments have no terminology at all.

![image](https://cloud.githubusercontent.com/assets/313942/20575272/98624a36-b186-11e6-97e4-0c4a986c2146.png)

@che85 can we assign Tissue/Tissue in a new segment event handler? We went over this several times to change the default behavior in Segment Editor proper, and never reached consensus.

As it is right now, the user needs to assign terminology manually before being able to export. My goal is to make user minimally aware of either terminology or DICOM, and not to introduce extra work.

Also, let's change the column name from "Color" to "Terminology", and assign segment name to be `<segmentation property type>_<segment number>` - I think this is more intuitive than `Segment_<segment number>`.

These are nice to have, but not critical for RSNA. Let me continue with the review of #102 and after I am done we will prioritize.

pieper:
Wouldn't the column name be "Term" instead of "Terminology"?

fedorov:
@pieper  -- agreed!

lassoan:
Renaming Color to Terminology would be confusing. Maybe we should bring up the terminology window when segmentation name column is double-clicked. It would be better if double-clicking on color swatch would bring up a simple color selector anyway (now you need to do too many clicks to simply change color).

fedorov:
@pieper - maybe it should be even more specific - maybe "semantics" is appropriate in this context?

fedorov:
@lassoan - why would it be confusing? I think it is more confusing when user only sees color, and has no idea that that color square is hiding more than just the color. To clarify, I am not pushing for a change in Segment Editor (although, it would be appropriate too), but only in this Reporting module.

che85:
I totally agree, but I also think that user just might not know about the fact that he/she is actually able to change the terminology by dblclicking onto the color. 

We shall find a better way for making that more visible more intuitive to the user.

fedorov:
@lassoan I also agree that changing color independently is a good feature. Maybe we should have separate column for color, and separate column/button for defining the semantics of the segment. That would be fine with me. All I am saying the current UI is sub-optimal. 

But we DO have a functional Beta, and all of the team contributed a lot to make it happen. Thank you all for the hard work!!!

lassoan:
We all agree then: Associating color with terminology is confusing.
Renaming column would be confusing because colors are very weakly linked to terminology and we would still just show colors in that column.

[...]

Issue #79, opened 2016-10-07
Segment name should be initialized from 3dSlicerLabel attribute in the terminology context


fedorov:

cpinter:
I can implement this feature easily, but I didn't do that on purpose. As I explained earlier:
"We also agreed that if a segment has a default name, then it is changed from the chosen terminology. I didn’t implement this feature, because some things are not clear: How do we assemble the name? (if the type has a modifier, or if we want to use 3dSlicerLabel then it may not be found, languages etc.) How do we make sure the user does not want the name that we identified as default? Again, it would probably be better to make sure some terminology is selected first, as the users should know what they are segmenting."

Using the 3dSlicerLabel field makes sense, but questions still remain.
Should I just ignore those and add this default name replacement feature?

fedorov:
Right, but also as I responded earlier.

> > ·         We also agreed that if a segment has a default name, then it is
> > changed from the chosen terminology. I didn’t implement this feature,
> > because some things are not clear: How do we assemble the name? (if the type
> > has a modifier, or if we want to use 3dSlicerLabel then it may not be found,
> > languages etc.) How do we make sure the user does not want the name that we
> > identified as default? Again, it would probably be better to make sure some
> > terminology is selected first, as the users should know what they are
> > segmenting.
> >
> 
> I don't have a solution about how to assemble a name. I think for now
> it makes sense to use "3dSlicerLabel" as that name. Where
> "3dSlicerLabel" is not populated, we can add a meaningful value. Where
> we have a modifier, we can augment the name.
> 
> If the user doesn't like the default, it will be editable, right?
> Changing the name of the segment doesn't change the codes for the
> terminology. It is just an uncontrolled free text.
> 
> Definitely the user should first select the terminology before doing
> any segmentation, and GeneralAnatomy should be that default selected
> automatically, I think.
> 
> I agree with you on almost all points, except I don't think default
> should be assigned at the time of storage.
> 
> I think user should be forced to choose the default, and be aware of
> what default was chosen. I also think that from the perspective of a
> user who doesn't want to deal with terminology, the impact should be
> minimal.
> 
> I think it is possible to achieve this rather easily. I believe in
> most cases GeneralAnatomy LUT is used as a default LUT. We can keep it
> the same way. The only thing that changes is that now each of the
> items in the LUT has a code assigned. The first item in that LUT is
> "Tissue", and I think it makes sense to keep that as default.
> 
[...]

PR #122, opened 2016-12-02
Cleanup temp directories, assigning initial terminology to new segments, and added slicelet

### LINKED ISSUE #116: Remove temporary files from DICOM plugins
- [x] DICOMParametricMapPlugin
- [x] DICOMSegmentationPlugin
- [x] DICOMTID1500Plugin
- [x] copy the DICOM files created by the converters to the DICOM DB directory

che85:
fixes #116 
addresses #104

@fedorov please test this. I hope that fixes the the issues regarding the missing files after restarting Slicer, even when I am not sure, why I cannot reproduce it properly.

### DIFF
--- a/Py/QuantitativeReporting.py
+++ b/Py/QuantitativeReporting.py
@@ -491,13 +491,10 @@ class QuantitativeReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidge
 
     if cliNode.GetStatusString() != 'Completed':
       raise Exception("tid1500writer CLI did not complete cleanly")
-    indexer = ctk.ctkDICOMIndexer()
-    indexer.addFile(slicer.dicomDatabase, outputSRPath)
 
   def addProducedDataToDICOMDatabase(self):
-    databaseDirectory = qt.QSettings().value("DatabaseDirectory")
     indexer = ctk.ctkDICOMIndexer()
-    indexer.addDirectory(slicer.dicomDatabase, self.tempDir, os.path.join(databaseDirectory, "dicom"))
+    indexer.addDirectory(slicer.dicomDatabase, self.tempDir, "copy")  # TODO: doesn't really expect a destination dir
 
   def cleanupTemporaryData(self):
     try:
--- a/Py/QuantitativeReporting.py
+++ b/Py/QuantitativeReporting.py
@@ -368,15 +368,14 @@ class QuantitativeReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidge
     defaults = ['T-D0050', 'SRT', 'Tissue']
     segmentTerminologyTag = terminologyWidget.serializeTerminologyEntry(
       loadedTermNames.GetValue(int(loadedTermNames.GetNumberOfValues()-2)),
-      defaults[0], defaults[1], defaults[2], "", "", "",
       defaults[0], defaults[1], defaults[2],
+      defaults[0], defaults[1], defaults[2],
+      "", "", "",
       loadedAnatNames.GetValue(int(loadedAnatNames.GetNumberOfValues()-1)),
       "", "", "", "", "", "")
     segment.SetTag(vtkSegmentationCore.vtkSegment.GetTerminologyEntryTagName(),
                    segmentTerminologyTag)
-
-    # TODO: if one segment 'Tissue_2' has been deleted and 'Tissue_1' and 'Tissue_3' are the only ones left and
-    # adding a new segment we have twice 'Tissue_3'
+    segment.SetColor(map(lambda c: float(c) / 255., [128,174,128]))  # TODO: get that value from context?
     segment.SetName("{}_{}".format(defaults[2], str(self.segmentEditorWidget.getNumberOfSegmentsStartingWith(defaults[2])+1)))
 
   @postCall(refreshUIElementsAvailability)
@@ -900,7 +899,7 @@ class CustomSegmentStatisticsLogic(SegmentStatisticsLogic):
       category = terminologyEntry.GetCategoryObject()
       segmentData["SegmentDescription"] = category.GetCodeMeaning() if category != "" else self.statistics[segmentID, "Segment"]
       segmentData["SegmentAlgorithmType"] = "MANUAL"
-      rgb = segment.GetDefaultColor()
+      rgb = segment.GetColor()
       segmentData["recommendedDisplayRGBValue"] = [rgb[0]*255, rgb[1]*255, rgb[2]*255]
       segmentData.update(self.createJSONFromTerminologyContext(terminologyEntry))
       segmentData.update(self.createJSONFromAnatomicContext(terminologyEntry))
--- a/Py/QuantitativeReporting.py
+++ b/Py/QuantitativeReporting.py
@@ -177,13 +177,10 @@ class QuantitativeReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidge
     self.selectionAreaWidgetLayout = qt.QGridLayout()
     self.selectionAreaWidget.setLayout(self.selectionAreaWidgetLayout)
 
-    self.loadDicomButton = qt.QPushButton("Load DICOM")
-
-    self.selectionAreaWidgetLayout.addWidget(self.loadDicomButton, 0, 0, 1, 2)
-    self.selectionAreaWidgetLayout.addWidget(qt.QLabel("Measurement report"), 1, 0)
-    self.selectionAreaWidgetLayout.addWidget(self.measurementReportSelector, 1, 1)
-    self.selectionAreaWidgetLayout.addWidget(qt.QLabel("Image volume to annotate"), 2, 0)
-    self.selectionAreaWidgetLayout.addWidget(self.imageVolumeSelector, 2, 1)
+    self.selectionAreaWidgetLayout.addWidget(qt.QLabel("Measurement report"), 0, 0)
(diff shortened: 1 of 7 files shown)


Py/QuantitativeReporting.py
```diff
@@ -317,6 +320,31 @@ def createNewSegmentationNode(self):
...
+      loadedTermNames.GetValue(int(loadedTermNames.GetNumberOfValues()-2)),
+      defaults[0], defaults[1], defaults[2], "", "", "",
+      defaults[0], defaults[1], defaults[2],
+      loadedAnatNames.GetValue(int(loadedAnatNames.GetNumberOfValues()-1)),
+      "", "", "", "", "", "")
+    segment.SetTag(vtkSegmentationCore.vtkSegment.GetTerminologyEntryTagName(),
+                   segmentTerminologyTag)
+
+    # TODO: if one segment 'Tissue_2' has been deleted and 'Tissue_1' and 'Tissue_3' are the only ones left and
+    # adding a new segment we have twice 'Tissue_3'
```
fedorov: It's ok - they can be edited

fedorov:
I first tested with an older nightly, and default terminology was not populated (some complaint in the popup). with today's nightly, I get the error below:

image

Do you get this error with the nightly?

fedorov:
That error was while trying to load an SR with plugin. While creating a new report and adding segments, I am still getting errors for initializing terminology for new segment:

image

In the terminal, I am getting this:

image

I confirm I am testing https://github.com/QIICR/QuantitativeReporting/pull/122/commits/aae5c30bbb6f3a7e734319c484cf51d41f5bd7d1

che85:
@fedorov I tested that with my nightly build from 11/26 but I am just downloading the most recent nightly for testing and fixing.

che85:
@cpinter Is it possible to get the Color of a category/type from a specific context?

@fedorov It should work now. There were some changes on the core...

cpinter:
qSlicerTerminologyNavigatorWidget::recommendedColorFromTerminology(vtkSlicerTerminologyEntry*)
Is this what you mean?

I suggest reading through qSlicerTerminologyNavigatorWidget.h to get an idea about the available utility functions.

che85:
@cpinter qSlicerTerminologyNavigatorWidget::recommendedColorFromTerminology is static but not  Q_INVOKABLE

cpinter:
Do you need Q_INVOKABLE for other methods as well? Please take a look and I'll add them.

che85:
I think that should be all for now.

cpinter:
Committed

che85:
@fedorov Please test and me know if everything works.

fedorov:
I didn't forget, just few a bit more urgent issues I need to work on.

che85:
No rush, I am cleaning up SliceTracker a bit

fedorov:
It looks very nice, with few suggestions:

 - when user changes terminology for segment, I think it would be more intuitive if the segment name was automatically updated to correspond to the new tissue property type
 - what do you think about popping up Terminology dialog when user requests a new segment creation, with tissue/tissue pre-selected in the dialog?

che85:
> when user changes terminology for segment, I think it would be more intuitive if the segment name was automatically updated to correspond to the new tissue property type

> what do you think about popping up Terminology dialog when user requests a new segment creation, with tissue/tissue pre-selected in the dialog?

I can implement that.

To me the popup makes sense and would only be one more click. The only drawback would be that we force the user to accept (select a terminology), which could also become annoying. In my opinion there should be a checkbox "don't display again" or something similar like "set as default".

fedorov:
> In my opinion there should be a checkbox "don't display again" or something similar like "set as default".

Yes, makes sense. I agree. Bit this would need to be set for the component outside of the module, so potentially problematic.

cpinter:
The popup with the default selection checkbox makes sense to me. The default selection should be per-segmentation I think, not application-wide.
What should happen with the color if the default is on? Consecutive colors from generic anatomy?

che85:
That sounds reasonable to me. 

@cpinter Another question that I have. Is there any event other than ModifiedEvent, that fires when a terminology has been set? I am asking that for handling segment names in our module which should be adapted to the terminology once that has been set. The segment name would only be automatically set when the terminology changes.

cpinter:
Currently only the Modified event is invoked (see vtkSegment::SetTag).
As terminology is stored in a tag, a TagAdded or TagModified (or similar) event could be the custom event that is fired without changing the underlying data structure. Or we could use a special member to store terminology, in which case a completely specific TerminologyChanged event could be used.

che85:
That would make sense to me.

cpinter:
I described two options. One is less specific (woud invoke with any tag setting) but much easier to implement.
@lassoan What do you think?

lassoan:
It would be difficult to add meaningful support for terminology in the low-level vtkSegment class. I would recommend to keep terminology info storage in tag(s) and add a TagModified custom event.

We'll have to implement default segment name and color support in the Slicer core anyway, so I think it's better not to spend time with a workaround but just let Csaba implement it in the core. The idea is to have two flags in vtkSegment that indicate if segment name and color have been automatically generated / the user specified them explicitly. If they are automatically generated then they are automatically updated when user selects a different terminology item.

@cpinter By when do you think this default segment name&color support could be implemented?

cpinter:
Even tomorrow if needed. @fedorov Please set a deadline for me and I'll make sure I meet it!

fedorov:
@cpinter - there is no urgency, I think it will be good to resolve these small issues before the project week, but there is no hard deadline, now that RSNA is over. But if you would like me to set a deadline -early next week would be good. I think the idea that Andras suggests is reasonable. Let's make sure those flags are accessible from python! ;)

che85:
I added a slicelet for quantitative reporting

image

cpinter:
We have discussed the terminology widget with Andras, and we came up with the following:
* Name can be changed from terminology widget (since it is populated based on that selection)
* Double-clicking the name would gring up the terminology selector, instead of directly editing the custom name (because both name and terminology answers the 'what is this' question, as opposed to color which is mostly a display property). Color would be changed with terminology eslection as before. One question here whether we pop up the terminology selector when double-clicking the color, or should it bring up a color selector for custom color?

We imagined the terminology selector this way:
image

What do you think?
I'll start implementing the changes after lunch if we can agree on the design and behavior.

fedorov:
> We imagined the terminology selector this way

Two suggestions:
* I would change the terminology selector width to extend over category/type lists only, since anatomic region context is controlled independently. 
* Earlier we discussed a checkbox to control whether terminology navigator would show up every time a new segment is added, or user is fine with just populating the default terminology/color values. Do you plan to add it? I don't see it in the redesigned widget.

Other than that, the design looks good.

> Double-clicking the name would bring up the terminology selector

I think it is rather important to be able to edit segment name without having to change the terminology for the segment. I would still prefer a separate button to invoke terminology navigator, rather than having to double-click on the segment name. 

> One question is whether we pop up the terminology selector when double-clicking the color, or should it bring up a color selector for custom color?

I think it is be more intuitive if clicking the color icon brings up color selection widget.

cpinter:
> I would change the terminology selector width to extend over category/type lists only, since anatomic region context is controlled independently
Sorry but I don't understand this fully. Can you please elaborate?
The anatomic region section would only show up if the selected category has showAnatomy on, otherwise the dialog would contain only the two (category and type).

If we consider terminology information necessary, then showing the terminology selector when clicking on name would make sense. We'd encourage the user to choose a meaningful terminology, but they could define custom name if necessary.
On the other hand if terminology information is optional, then changing the name immediately when double-clicking the name item is OK.
I think this is the main question.

fedorov:
>> I would change the terminology selector width to extend over category/type lists only, since anatomic region context is controlled independently
>Sorry but I don't understand this fully. Can you please elaborate?

Here's what I meant (roughly!)

image

For the segment list, I meant to suggest the following:

image

cpinter:
Thanks for the sketches! It makes sense, I already changed the widget. I'll implement it that way.

The segment table behavior change is independent from my current work on the default segment name, so I'll not change it yet and let it ripen a bit.
I personally think that showing terminology information in a separate cell would make it too wide (just consider how many separate fields we can have: type, modifier, anatomic region, modifier), so I'd keep it in a tooltip.

Should we move this discussion to its ticket https://github.com/QIICR/QuantitativeReporting/issues/79 ?

fedorov:
> The segment table behavior change is independent from my current work on the default segment name, so I'll not change it yet and let it ripen a bit.
I personally think that showing terminology information in a separate cell would make it too wide (just consider how many separate fields we can have: type, modifier, anatomic region, modifier), so I'd keep it in a tooltip.
Should we move this discussion to its ticket #79 ?

Yes, sure - no urgency. I suggest we sit down at the project week to discuss options and develop consensus.

lassoan:
A. If you want to show terminology info in the segment list and there is enough space then it could be good to have a terminology column. Terminology popup can be shown by double-clicking terminology item.

B. If there are not so much space (or terminology information is not important to show) then it's better to not show terminology column. Terminology popup can be then shown by double-clicking segment name. Note that no extra clicks are needed to change segment name using terminology popup (focus can be set by default on the segment name widget, Enter closes the popup).

Both A and B make sense, in different use cases. So, probably the behavior should be configurable.

fedorov:
> probably the behavior should be configurable

👍

cpinter:
Change about initial terminology committed in
https://github.com/Slicer/Slicer/commit/00f4d75d9695a97d5926fe073c48077449130829

fedorov:
@cpinter thank you!

Issue #120, opened 2016-11-27
RSNA2016 bug list


fedorov:
Let's keep adding here issues identified during the conference instead of creating a new issue each time. After RSNA we can triage and decide how to proceed.

- [x] the module does not work in Slicer installed from scratch, since developer mode is not enabled, and it cannot find widgets!
- [x] SR round-trip does not work if Slicer is restarted - somehow the DICOM files created by the SR converter are not available after restart (either temp dir is removed by the OS, or something else; probably the files created by the converters should be moved to the DICOM import directory)
- [x] when switching to 4up-table layout, the table in the layout is not responsive to the mouse clicks; perhaps this is intentional, but not intuitive
- [x] for SRs that reference RWVM, scaling should be applied (specifically in the PET/SUV scenario)

lassoan:
"it cannot find widgets" => In general, only testing code should rely on finding widgets by Qt meta-object system, as these kind of lookups are invisible to developers and routine code changes may accidentally break them. It should be no problem to add the proper APIs to access all necessary widgets without searching.

lassoan:
> switching to 4up-table layout, the table in the layout is not responsive to the mouse clicks; perhaps this is intentional, but not intuitive

Can you reproduce it by simply loading/creating table using Tables module and switching to 4up-table layout?

Table node supports locking to prevent accidental modifications. If a table is locked then you can still browse and copy the content but cannot modify it.

fedorov:
> "it cannot find widgets" => In general, only testing code should rely on finding widgets by Qt meta-object system

I agree. Let's explore this. This approach is great for prototyping, but is not sustainable. Let's see how easy it will be to add all of the API calls to support this functionality properly.

> Can you reproduce it by simply loading/creating table using Tables module and switching to 4up-table layout?

Andras, I think this is related to how we customized table events for this module. What I asked Christian to implement is synchronized selection of rows in the table with the selection of segments in the segments list (this is the PR I believe: https://github.com/QIICR/QuantitativeReporting/pull/112/commits/98b2b85377b2dafc38cfab99f705b31093c896fb). But somehow the behavior is not the same when layout is changing. Maybe it changes the view in the 4up-table layout, and the event handler does not apply? @che85 what do you say?

che85:
@fedorov I totally forgot about the 4up-table layout. Generally the module is only listening to click events but I should rather let it listen to selection changed events.

[...]

PR #126, opened 2017-01-19
ENH: account for the referenced RWVM instance

fedorov:
re #120

When measurements SR references a RWVM instance, check whether the series instance
UID referenced by SEG is the same as the one referenced by RWVM. If yes, use RWVM
plugin to load the scaled series.

This does not add support for saving segmentation of volumes loaded from RWVM.

### DIFF
--- a/Py/DICOMTID1500Plugin.py
+++ b/Py/DICOMTID1500Plugin.py
@@ -11,6 +11,7 @@ class DICOMTID1500PluginClass(DICOMPluginBase):
   UID_EnhancedSRStorage = "1.2.840.10008.5.1.4.1.1.88.22"
   UID_ComprehensiveSRStorage = "1.2.840.10008.5.1.4.1.1.88.33"
   UID_SegmentationStorage = "1.2.840.10008.5.1.4.1.1.66.4"
+  UID_RealWorldValueMappingStorage = "1.2.840.10008.5.1.4.1.1.67"
 
   def __init__(self):
     super(DICOMTID1500PluginClass, self).__init__()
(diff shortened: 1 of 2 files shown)


che85:
@fedorov Which dataset did you use for testing?

fedorov:
I used one of the studies from QIN-HEADNECK

che85:
@fedorov when using QIN-HEADNECK-01-0139 and trying to load the SR I get the following error.

```
Traceback (most recent call last):
  File "/Applications/Slicer.app/Contents/lib/Slicer-4.7/qt-scripted-modules/DICOMLib/DICOMWidgets.py", line 790, in loadCheckedLoadables
    self.proceedWithReferencedLoadablesSelection()
  File "/Applications/Slicer.app/Contents/lib/Slicer-4.7/qt-scripted-modules/DICOMLib/DICOMWidgets.py", line 840, in proceedWithReferencedLoadablesSelection
    if not plugin.load(loadable):
  File "/Users/christian/Documents/SPL/sources/py/Reporting/Py/DICOMTID1500Plugin.py", line 134, in load
    rwvmFile = slicer.dicomDatabase.filesForSeries(loadable.ReferencedRWVMSeriesInstanceUIDs[0])[0]
IndexError: tuple index out of range
```

fedorov:
Interesting, thanks. Can you debug why this is happening (failure to look up the file corresponding to the referenced RWVM)? In any case, of course the code should be robust to this situation. I should fix it.

che85:
yes I can look into that

che85:
@fedorov the SeriesInstanceUID referenced within the SR doesn't match the one delivered with the datasets. I downloaded it from here http://slicer.kitware.com/midas3/folder/3771

fedorov:
Can you download patient 139 from TCIA and check if the problem exists there as well?

che85:
Downloading patient 139 directly from TCIA delivers the right RWV series with the exact same SeriesInstanceUID that is missing on midas3

che85:
So yes it works fine

fedorov:
ok, thanks! I will improve the code and re-upload the correct RWVM to Midas. Can you attach the correct RWVM instance to this issue to make my life simpler?

che85:
[000000.dcm.zip](https://github.com/QIICR/QuantitativeReporting/files/720140/000000.dcm.zip)

fedorov:
where exactly did you download the data from midas? I don't see RWVM here at all: http://slicer.kitware.com/midas3/folder/3854

fedorov:
ah, I see! I think what happened you imported PET using DICOM RWVM plugin (the date on your screenshot is today!), it was never downloaded from Midas. So I will just create a new folder there for RWVM. Thanks!

Py/DICOMTID1500Plugin.py
```diff
@@ -127,19 +127,22 @@ def load(self, loadable):
         if hasattr(segLoadable, "referencedSeriesUID"):
           # determine if RWVM needs to be applied to the series referenced from SEG
           loadRWVM = False
-          if len(loadable.ReferencedRWVMSeriesInstanceUIDs)>0:
+          rwvmUIDs = loadable.ReferencedRWVMSeriesInstanceUIDs
+          if len(rwvmUIDs)>0 and len(slicer.dicomDatabase.filesForSeries(rwvmUIDs[0]))>0:
```
che85: I wouldn't run filesForSeries that often since there will always be executed another query on the dicomDatabase [1]. I would reuse the value after having run it once

[1] https://github.com/commontk/CTK/blob/04607d0ea7fbab104ede38c7c2acc7b5dd1fdfa0/Libs/DICOM/Core/ctkDICOMDatabase.cpp#L786-L796

fedorov: I agree, good point!

pieper:
It's true that it's another query, so saving the result is good if you plan
to use it again anyway.  But the query should be fast because the Images
table is indexed on SeriesInstanceUID.

https://github.com/commontk/CTK/blob/04607d0ea7fbab104ede38c7c2acc7b5dd1fdfa0/Libs/DICOM/Core/Resources/dicom-schema.sql#L72

Do you observe any performance issues with the call filesForSeries call?

che85:
@pieper No I didn't notice any performance issues with it. I only noticed, that it doesn't need to be called more than once.

fedorov:
@che85 let me know if you have any other concerns, or merge otherwise

PR #139, opened 2017-01-31
Gitbook documentation update

che85:
It's still wip but you can already merge. I will continue working on this later tomorrow (today) and add missing screenshots.


fedorov:
Thank you! I think I will do the edits directly in the text, this will be more efficient, but I won't get to it before Thu/Fri probably.

Issue #143, opened 2017-02-23
API to better handle visibility of individual UI elements


fedorov:

cpinter:
As far as I know @lassoan took this task, please assign to him

che85:
The following components should be directly accessible from python:
* (maybe) UndoRedoButtonBox
  * UndoButton
  * RedoButton,
* EffectsGroupBox
* OptionsGroupBox
* MaskingGroupBox
* AddSegmentButton
* RemoveSegmentButton
* CreateSurfaceButton
* segment table and the status (label?) on top of it.

It might make sense to hold a list of UI component object names which should be accessible by developers.

```
getAccessibleUIComponents()
```

We could then easily access those components by calling 
```
getUIComponentByObjectName(..)
```
For being able to configure available effects I suggest to add something like 

```
setAvailableEffects([Paint, Draw, Erase, ...])
```

che85:
@cpinter Sorry I wasn't sure.

lassoan:
I would prefer to control what developers can change in the widget to make sure we can keep the editor widget appearance and behavior consistent across all modules (and if you have good ideas for improvements then not only Reporting but all other modules should use that). For example, we could develop maybe 2-3 alternative layouts (normal, space-saving, horizontal) with a few options (show/hide buttons and sections), instead of just letting developers move controls around randomly (that would be hard to control and maintain).
Could you write what are the specific changes that you need in the appearance or behavior?

setAvailableEffects([Paint, Draw, Erase, ...]) is a separate story, I'll add a new ticket for that.

fedorov:
@che85 can you please respond?

che85:
@lassoan Why not exposing all components? I as a developer want to be free in how to organize the UI elements. I don't agree with having predefined layout unless you have a designer who did some research on how to propose components the best way to users.

A UI should always be space-saving but highlighting the important components for guaranteeing the best possible user experience. If you want developers to reuse your components, you should either give them freedom in how to organize the UI.

For QuantitativeReporting I just want to be as flexible as possible and hide components of the SegmentEditor that I don't want to get exposed to the user.

Who would define how normal, space-saving and horizontal will look like?

lassoan:
> don't agree with having predefined layout unless you have a designer

GUIs are always problematic, as it is difficult to gather hard evidence and everybody has his own opinion. Don't expect anything from UX designers, unfortunately more often they make more harm than good.

[...]

PR #154, opened 2017-04-05
ENH: Adaption to changes/improvements in API for accessing SegmentEditor UI elements

### LINKED ISSUE #153: Loading SR with multiple referenced series
loads only segmentation and measurements, but no referenced series

reported by @chribaue

che85:
* added icon for displaying in Slicer module selector
* cleanup directories and deleted unused files

fixes #153
working on #143

depends on https://github.com/Slicer/Slicer/pull/699


Plugins/DICOMParametricMapPlugin.py
```diff
@@ -1,164 +1,165 @@
-import os
```
fedorov: I think it would be good to call this directory `DICOMPlugins` instead of `Plugins`

fedorov:
We should spend more time debugging this.

I am also annoyed by this message below that is not 100% reproducible. Can you try loading QIN-HEADNECK subject 24? That's where it failed for me...

image

fedorov:
The error I listed above is reproducible with the "master", so it is not the fault of this PR.

fedorov:
Maybe we should not rely on referenced instances passed to DICOMWidget, and load referenced items individually by calling the corresponding plugins from the TID1500 plugin.

fedorov:
@pieper can you look into this? I think the failure above may be related to the recent change in dealing with the plugins that generate warnings. Our plugins worked until recently, and nothing changed on our side. I think you understand this function better than me: https://github.com/Slicer/Slicer/blame/master/Modules/Scripted/DICOMLib/DICOMWidgets.py#L805

fedorov:
@pieper for the warning regarding "71: tumor measurements" loadable above - it appears that this is happening because DICOMWidget prints warnings/examines loadables that are not even checked, in this case a loadable produced by Scalar volume plugin for an SR series. Is this really the expected behavior?

fedorov:
As it turns out, I was testing with a faulty dataset. Error checking needs to be improved in the plugin. When segmentations are not loaded, it will fail. Christian, I won't have time to work on this until after April 18. Please debug the issues we observed during the call today. The expected behavior should be as in the screenshot below (from the 3/20/17 nightly)

image

pieper:
Do you still need input from me?

fedorov:
@pieper there were two issues where would be great to get your input:

1) it looks like loadables that are unchecked trigger warnings during load process. Is this expected?
2) in some situations (@che85 will follow up with details) we were getting DWI plugin generating a popup message with a warning, even though the series only contained PET/CT/SEG/SR series. Not clear why it was happening. Would you expect this kind of behavior?

pieper:
1: no, not expected.  The examine step [1] should only be happening on
plugins that are selected (box checked [2])

2: the diffusion plugin uses a heuristic since DWI scans come in many
different formats.  I think the best thing to do would be to add some
trivial reject code to [3], for example there's no way that a PET/CT/SEG/SR
SOPClassUID or Modality would be used to represent a DWI.  @che85 if you
are following up with details on reproducing this can you put it in the
form of a slicer bug report and then we can come up with a robust
workaround.

[1]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMLib/DICOMWidgets.py#L693

[2]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMLib/DICOMWidgets.py#L1038

[3]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMPlugins/DICOMDiffusionVolumePlugin.py#L81-L100

lassoan:
We keep having problems with DWI plugin interfering with series loading. As SlicerDiffusion is an extension now, I think we should move the DWI plugin into the extension. It would make DICOM parsing faster and more robust for users who don't use diffusion imaging. @che85 could you ask Isaiah Norton about this?

pieper:
I agree, the false positives are annoying so moving the DWI plugin to the
extension sounds like a good step.  For the record, the DWI plugin only
looks at a few tags in one file so it's probably not a performance
bottleneck.

fedorov:
I think moving DWI plugin to an extension is a separate decision. What seems to be happening is that warnings generated for loadables that are not checked get printed out and perhaps even show up in popups, which is confusing. It is more important to confirm this is happening, and if yes - address this issue in DICOMWidgets.py.

pieper:
Do you have a simple way to replicate the issue?

che85:
Seems like there are some new issues in the most recent nightly build.

I cleaned up the DICOMDatabase and freshly imported QIN-HEADNECK-01-0024. Loading the following SR

image

I get this popup:

image

and error message:

```
ReadData: Cannot read file as a volume of type Volume[fullName = /Users/christian/Documents/QIN-HEADNECK/QIN-HEADNECK-01-0024/1.3.6.1.4.1.14519.5.2.1.2744.7002.271803936741289691489150315969/1.2.276.0.7230010.3.1.3.8323329.20938.1440001386.735918/000000.dcm]
	Number of files listed in the node = 1.
	File reader says it was able to read 0 files.
	File reader used the archetype file name of /Users/christian/Documents/QIN-HEADNECK/QIN-HEADNECK-01-0024/1.3.6.1.4.1.14519.5.2.1.2744.7002.271803936741289691489150315969/1.2.276.0.7230010.3.1.3.8323329.20938.1440001386.735918/000000.dcm []
ITK exception info: error in unknown
 Could not create IO object for reading file /Users/christian/Documents/QIN-HEADNECK/QIN-HEADNECK-01-0024/1.3.6.1.4.1.14519.5.2.1.2744.7002.271803936741289691489150315969/1.2.276.0.7230010.3.1.3.8323329.20938.1440001386.735918/000000.dcm
  Tried to create one of the following:
    NiftiImageIO
    NrrdImageIO
    GiplImageIO
    JPEGImageIO
    GDCMImageIO
    BMPImageIO
    LSMImageIO
    PNGImageIO
    TIFFImageIO
    VTKImageIO
    StimulateImageIO
    BioRadImageIO
    MetaImageIO
    MRCImageIO
    MGHImageIO
    MRMLIDImageIO
    HDF5ImageIO
    GE4ImageIO
    GE5ImageIO
  You probably failed to set a file suffix, or
    set the suffix to an unsupported type.
```

and 

```
Warning in DICOM plugin Scalar Volume when examining loadable 93: tumor measurements - User3 SemiAuto trial 2: There is no pixel data attribute for the DICOM objects, but they might be readable as secondary capture images.  Reference image in series does not contain geometry information.  Please use caution.  

Could not load: 93: tumor measurements - User3 SemiAuto trial 2 as a Scalar Volume
```

Nothing gets loaded into Slicer.

che85:
@ihnorton What do you think?

> We keep having problems with DWI plugin interfering with series loading. As SlicerDiffusion is an extension now, I think we should move the DWI plugin into the extension. It would make DICOM parsing faster and more robust for users who don't use diffusion imaging. @che85 could you ask Isaiah Norton about this?

ihnorton:
I can move the DWI plugin, but this doesn't look like a DWI problem. For the first message, IIUC there should never be any ITK interaction when examining because ctkDICOMDatabase uses DCMTK directly, and then the python code just access the SQLite database. The second message is from the Scalar Volume plugin.

I tried the dataset `tumor_measurements_User1_Manual_Trial1.dcm` and did not get any errors from Examine, with both Diffusion Plugin and all QR extension plugins enabled.

che85:
@fedorov Loading "71: tumor measurements" with **master** works fine for me too on the current nightly build from 2017-04-09

image

che85:
I might have found a bug in this branch that I will need to investigate.

che85:
Here is something I don't quite understand when adding modules/extensions to Slicer:

Earlier, I refactored the directory structure of QuantitativeReporting in a manner that I have the main QuantitativeReporting module in the top level directory. Furthermore I put all DICOM plugins into a sub directory called `DICOMPlugins`. Both, the top level as well as `DICOMPlugins` directory hold a CMakeLists.txt where the top level CMakeLists.txt adds `DICOMPlugins` as a subdirectory. 

What I would expect when adding the top level of QuantitativeReporting to Slicer:
* QuantitativeReporting is available in module selector 👍 
* all underlying DICOMPlugins should be available (but none of them is unless I add DICOMPlugins explicitly as to additional module paths) 👎 

Obviously it works when I build the extension and then include the qt-scripted-modules from the build/lib/Slicer4.7 directory

I there anything I am missing?

che85:
Ok for development purpose it's ok to add two directories, but in general Slicer should also look in sub directories... 

When building the QuantitativeReporting, the extension and DICOMPlugins are anyway in the same directory, so it should work just from installing it from the Slicer Extension Manager.

che85:
@fedorov I think we should merge this PR and go from there since it is not only related to QuantitativeReporting. This PR includes changes that adapt to the SegmentEditor improvements.

fedorov:
Yes, ok.

lassoan:
There are a few simple rules:
* Slicer will attempt to load all .py files that are placed in a module directory as a Slicer module. Therefore, all .py files should contain a module class in these .py files that have the same name as the .py file. The module can be hidden to prevent it showing up in the module list.
* To be able to test your module by simple adding top-level directories to the module path, you must use the same directory structure in the source tree as in the install tree. Don’t add CMakeLists.txt files in subdirectories, just make subdirectories proper Python modules (with. __init__.py file) and load it from your Slicer modules when needed.

There are several working examples, such as https://app.assembla.com/spaces/slicerrt/subversion/source/HEAD/trunk/SlicerRt/src/DicomRtImportExport

> general Slicer should also look in sub directories

We would lose the ability to define Python modules (directory containing __init__.py and various other .py files) if Slicer look for scripted loadable modules in subdirectories. Therefore, it is a feature, not a limitation, that Slicer only attempts to load scripted loadable modules from top-level directories.

che85:
Thanks @lassoan for the comments on that. I will keep that in mind and improve the code here.

che85:
When using examine directly from the DICOMBrowser I don't get the DiffusionPlugin as the first priority reader:

image

When loading it with the DICOMTID1500Plugin I get the following popup:
image

So it's most probably related to our DICOMTID1500Plugin and the way we are loading the references.

> 1: no, not expected.  The examine step [1] should only be happening on
plugins that are selected (box checked [2])

>2: the diffusion plugin uses a heuristic since DWI scans come in many
different formats.  I think the best thing to do would be to add some
trivial reject code to [3], for example there's no way that a PET/CT/SEG/SR
SOPClassUID or Modality would be used to represent a DWI.  @che85 if you
are following up with details on reproducing this can you put it in the
form of a slicer bug report and then we can come up with a robust
workaround.

>[1]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMLib/DICOMWidgets.py#L693

>[2]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMLib/DICOMWidgets.py#L1038

>[3]
https://github.com/Slicer/Slicer/blob/5e6f95f524dfd70252326b589d17750300fc7d08/Modules/Scripted/DICOMPlugins/DICOMDiffusionVolumePlugin.py#L81-L100

che85:
@fedorov disabling the RWVM the following is loaded:

image

che85:
I think we should anyway add a check to the DICOMPlugin base class that checks if a series has already been added to Slicer. What do you think?

lassoan:
What would you do if the series has been already loaded?

fedorov:
@che85 I think adding this check as a utility function might be useful, but we cannot change the behavior in the general-purpose components of Slicer. You can never anticipate what the user wants to accomplish, so the whole idea of Slicer is to allow all those options and give flexibilty for some, at the cost of confusion for (probably, many) others.

I think the way to deal with this is at the level of a workflow/module. Inside QuantitativeReporting, we can force "scene close" equivalent on loading a new SR. But I also need to talk with you to understand better what is causing the issue, and why this is a problem.

che85:
@lassoan There could be a popup that notifies the user about an already loaded series with a checkbox "Don't show again" or something similar. 

@fedorov I can see that there are use cases where you explicitly want to load the same series several times but I can also see situations where you want to prevent that from happening and that's where at least a notification for the user/developer would be helpful. But other than that you are right.

lassoan:
It's difficult to guess what the user wants. Showing a popup disrupts the workflow. "Don't show again" does not solve the problem because the same user sometimes wants to have the same volume in the scene multiple times, while other times not.

Maybe having a checkbox in the DICOM dialog (near Advanced, Horizontal, Browser persistent) to ignore already loaded series would work.

PR #165, opened 2017-06-28
BUG: watchbox could not get populate due to missing storageNode

che85:
* now using DICOM.instanceUIDs instead.

depends on QIICR/SlicerDevelopmentToolbox/pull/9

### DIFF
--- a/QuantitativeReporting.py
+++ b/QuantitativeReporting.py
@@ -385,11 +385,14 @@ class QuantitativeReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidge
       segNode.AddObserver(vtkSegmentationCore.vtkSegmentation.SegmentAdded, self.onSegmentAdded))
 
   def initializeWatchBox(self, node):
-    try:
-      dicomFileName = node.GetStorageNode().GetFileName()
-      self.watchBox.sourceFile = dicomFileName if os.path.exists(dicomFileName) else None
-    except AttributeError:
+    if not node:
       self.watchBox.sourceFile = None
+      return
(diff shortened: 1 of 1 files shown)


QuantitativeReporting.py
```diff
@@ -385,11 +385,14 @@ def setupSegmentationObservers(self):
       segNode.AddObserver(vtkSegmentationCore.vtkSegmentation.SegmentAdded, self.onSegmentAdded))
 
   def initializeWatchBox(self, node):
-    try:
-      dicomFileName = node.GetStorageNode().GetFileName()
```
fedorov: @pieper this line returns `None` now, but used to work before. @che85 implemented a workaround, but I wonder if  this regression could be connected to the recent re-work of the DICOM readers and if yes perhaps it should be fixed on the Slicer side?

QuantitativeReporting.py
```diff
@@ -385,11 +385,14 @@ def setupSegmentationObservers(self):
...
-      self.watchBox.sourceFile = dicomFileName if os.path.exists(dicomFileName) else None
-    except AttributeError:
+    if not node:
+      self.watchBox.sourceFile = None
+      return
+    dicomFileName = slicer.dicomDatabase.fileForInstance(node.GetAttribute("DICOM.instanceUIDs").split(" ")[0])
```
fedorov: @che85 it is not impossible that this attribute is also missing. If that is the case, it means the volume was not loaded from DICOM, and there should be error communicated to the user.

che85: If the volume was not loaded from DICOM it wouldn't appear in the volume selector as we defined it earlier.

https://github.com/QIICR/QuantitativeReporting/blob/master/QuantitativeReporting.py#L173

Issue #140, opened 2017-02-14
DICOM export plugin


fedorov:
Currently it is only possible to export segments that were created within QuantitativeReporting. Need to investigate the generic DICOM export plugin mechanism to make this more broadly available. The challenge is that we will only be able to support export of segments done on a DICOM series.

TODO:
 - [x] test Subject Hierarchy 2.0 and look into how DICOM export plugins are supported there

fedorov:
TODO: 

 - test Subject Hierarchy 2.0 and look into how DICOM export plugins are supported there

cpinter:
The DICOM plugins work exactly the same way, but instead of storing subject hierarchy nodes, it stores subject hierarchy item IDs (and the getter functions etc are different). So if you check it out in the current Slicer code, the idea will be the same, and the code will be very similar.

The only thing I suggest is to wait for the new one before making any changes.

cpinter:
I came across SEG export, and I see it hasn't been updated yet. @fedorov do you want me to get started on porting it to SH2?

fedorov:
@cpinter yes this would be great!

cpinter:
I'm almost done, but there is one missing component, which is the CLI that does the export. It was called encodeSeg, and it was included in Reporting, but I cannot find it now. Can you please help me finding it?

fedorov:
@cpinter the converter CLI is now provided by the DCMQI extension, it is documented here: https://qiicr.gitbooks.io/dcmqi-guide/user_guide/itkimage2segimage.html, and example usage from Slicer is here: https://github.com/QIICR/QuantitativeReporting/blob/master/QuantitativeReporting.py#L501.

@che85 can you please work with Csaba to help him integrate the converter and pass all metadata properly? Please let me know when you think it is ready so I can review before it is merged!

cpinter:
Thanks, Andrey!

Probably it would make sense to create the PR now, because I'm at the point where SH2 and Terminologies are properly used, and the terminology triplet strings are created, and it's just the last few lines of the function that doesn't work, which is the one that calls the CLI.

Having written this down, it seems that this export plugin weren't updated several times when changes happened in the core and in DCMQI/QuantitativeReporting, each of which would have broken it itself. I think an automated test would be very useful!

che85:
I will take a look and submit a PR to @cpinter

che85:
[...]

PR #171, opened 2017-07-28
BUG: Fixed DICOM SEG plugin WIP

cpinter:
Calling the exporter CLI is not yet fixed

Fixes https://github.com/QIICR/QuantitativeReporting/issues/140

### DIFF
--- a/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -7,7 +7,6 @@ import logging
 #
 # This is the plugin to handle translation of DICOM SEG objects
 #
-
 class DICOMSegmentationPluginClass(DICOMPluginBase):
 
   def __init__(self):
@@ -25,7 +24,6 @@ class DICOMSegmentationPluginClass(DICOMPluginBase):
     return loadables
 
   def examineFiles(self,files):
-
     """ Returns a list of DICOMLoadable instances
     corresponding to ways of interpreting the
     files parameter.
@@ -210,7 +208,7 @@ class DICOMSegmentationPluginClass(DICOMPluginBase):
           labelNode.SetAttribute("ColorG", str(rgb[1]))
           labelNode.SetAttribute("ColorB", str(rgb[2]))
 
-          # create Subject hierarchy nodes for the loaded series
+          # Create subject hierarchy for the loaded series
           self.addSeriesInSubjectHierarchy(loadable, labelNode)
 
       metaFile.close()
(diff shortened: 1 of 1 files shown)


che85:
Should I also add checks if the segments of a segmentation have any actual pixeldata? I don't think that it should be possible to export empty segments into a DICOM segmentation.

fedorov:
@che85 looks like you are talking about 2 different issues:
1. SEG object does not have PixelData element at all (invalid object)
2. SEG object that has PixelData, but not all of the segments included in the object are populated (a segment could have 0 frames, or a segment could have non-zero frames, but all pixels set to 0 in its frames)

1 is trivial to check at the examine(), 2 could be checked after running the converter in load() and throw the error then.

fedorov:
After talking with @che85 I realized I misunderstood this was in the context of export, not import.

@cpinter do you have any recommendations how error-checking should be done at the time of export? What do you do if, for example, the user wants to export segmentation as RT, but some segments are empty? The behavior of the SEG exporter should probably be consistent with that of the RT export plugin.

cpinter:
There is no such check in SlicerRT, only geometry checks. In Plastimatch the structure is eventually added using this function:
https://gitlab.com/plastimatch/plastimatch/blob/master/src/plastimatch/base/segmentation.cxx#L244
I cannot see any check here either. Probably the empty structure is exported, and will trigger a warning when loading. Do you want me to check to make sure?

fedorov:
@cpinter thanks for checking! 

I think for the SEG plugin we should just detect the situation when a segment has no content, skip importing it, and give a warning to the user, unless you have a better idea.

cpinter:
I agree. I'll ask what Greg (he owns Plastimatch) thinks just in case.

fedorov:
for the record, the related discussion is below




---------- Forwarded message ----------
From: Andrey Fedorov
Date: Thu, Aug 17, 2017 at 2:13 PM
Subject: Re: Seg plugin fix
To: Andras Lasso
Cc: "Sharp, Gregory C.", Csaba Pinter, Christian Herz


> why would you remove an empty segment on export that the user intentionally kept?

Because I don't have an unambiguous and standard way to communicate to
the consumer of the object why I kept the empty segment in, what it
means, and what to do with it.

fedorov:
@cpinter, as I was re-reading this, to your comment, sorry I missed it earlier.

> You said you want to conform with how SlicerRT (and so Plastimatch) works

In my comment above I meant to conform to whatever warning/error notifications that I thought are present in SlicerRT. Since there are none, we will make one up for SEG export. Sorry for the confusion.

PR #174, opened 2017-08-11
BUG: fix renamed method in Slicer Core

che85:
for reference https://github.com/Slicer/Slicer/commit/37bb711bf7951ba48658a86faa0466d724d88afd


cpinter:
Sorry for not considering the QuantitativeReporting implications of this change! Thanks for the fix.

che85:
No problem :)

PR #175, opened 2017-08-16
ENH: added widgets for importing from segmentation/labelmap (issue #173)

### LINKED ISSUE #173: Initiallize quantitative report with an existing segmentation
It will be helpful if an existing segmentation could be attached (imported?) into a new quantitative report. It can be limiting to only be able to create segmentation from within the QR UI.

che85:
fixes #173 

image

depends on https://github.com/QIICR/SlicerDevelopmentToolbox/pull/17
depends on https://github.com/QIICR/SlicerDevelopmentToolbox/pull/18

### DIFF
--- a/QuantitativeReporting.py
+++ b/QuantitativeReporting.py
@@ -15,7 +15,8 @@ import vtkSegmentationCorePython as vtkSegmentationCore
 from SlicerDevelopmentToolboxUtils.mixins import ModuleWidgetMixin, ModuleLogicMixin, ParameterNodeObservationMixin
 from SlicerDevelopmentToolboxUtils.decorators import onExceptionReturnNone, postCall
 from SlicerDevelopmentToolboxUtils.helpers import WatchBoxAttribute
-from SlicerDevelopmentToolboxUtils.widgets import DICOMBasedInformationWatchBox
+from SlicerDevelopmentToolboxUtils.widgets import DICOMBasedInformationWatchBox, ImportLabelMapIntoSegmentationWidget
+from SlicerDevelopmentToolboxUtils.widgets import CopySegmentBetweenSegmentationsWidget
 from SlicerDevelopmentToolboxUtils.constants import DICOMTAGS
 from SlicerDevelopmentToolboxUtils.buttons import RedSliceLayoutButton, FourUpLayoutButton, FourUpTableViewLayoutButton
 from SlicerDevelopmentToolboxUtils.buttons import CrosshairButton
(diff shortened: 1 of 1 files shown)


che85:
The only thing that I don't like is that vtkSegmentation::CopySegmentFromSegmentation(..)[1] neither copies segment name nor does it set the same terminology.

[1] http://apidocs.slicer.org/master/classvtkSegmentation.html#a001c13d90bea679cb7697e33e56b59e5

lassoan:
> The only thing that I don't like is that vtkSegmentation::CopySegmentFromSegmentation(..)[1] neither copies segment name nor does it set the same terminology

All segment metadata (name, terminology, etc) should be copied. I've test it on a simple example and it works well for me:

<pre>
import vtkSegmentationCorePython as vtkSegmentation

# Create s1 segmentation with one segment
s1 = vtkSegmentation.vtkSegmentation()
s1seg1 = vtkSegmentation.vtkSegment()
s1seg1.SetName("sn1seg1name")
s1.AddSegment(sn1seg1)

# Create s2 segmentation
s2 = vtkSegmentation.vtkSegmentation()
# Copy and move segment from s1
s2.CopySegmentFromSegmentation(s1, s1.GetNthSegmentID(0), False)
s2.CopySegmentFromSegmentation(s1, s1.GetNthSegmentID(0), True)

# Printe segment names
print(s2.GetNthSegment(0).GetName())
print(s2.GetNthSegment(1).GetName())
</pre>

Can you give an example where segment metadata is not copied by CopySegmentFromSegmentation?

che85:
@lassoan I don't know what's going on. When moving/copying segments between segmentations after using the SegmentEditor only, it works. But when using QuantitativeReporting and copying from the segmentation (created with the SegmentEditor) to a segmentation created with QuantitativeReporting, the imported segments are renamed to Tissue_... and terminology is resetted.

che85:
One other thing that I noticed is when adding new segments in QuantitativeReporting, they always have the same initial color. Doing the same in SegmentEditor, the color is always different...

lassoan:
Are all Segmentation nodes added to the scene and scene is set in all MRML widgets?

che85:
I noticed that I had some obsolete code that was defining a default terminology for every added segment. That code existed before the terminology logic did that on its own.

che85:
added some icons for the buttons

image

QuantitativeReporting.py
```diff
@@ -15,7 +15,8 @@
 from SlicerDevelopmentToolboxUtils.mixins import ModuleWidgetMixin, ModuleLogicMixin, ParameterNodeObservationMixin
 from SlicerDevelopmentToolboxUtils.decorators import onExceptionReturnNone, postCall
 from SlicerDevelopmentToolboxUtils.helpers import WatchBoxAttribute
-from SlicerDevelopmentToolboxUtils.widgets import DICOMBasedInformationWatchBox
+from SlicerDevelopmentToolboxUtils.widgets import DICOMBasedInformationWatchBox, ImportLabelMapIntoSegmentationWidget
+from SlicerDevelopmentToolboxUtils.widgets import CopySegmentBetweenSegmentationsWidget
 from SlicerDevelopmentToolboxUtils.constants import DICOMTAGS
```
fedorov: consider using pydicom dictionary instead

PR #184, opened 2017-08-23
WIP: exporting to user selected directory and configuring availability of import area

### LINKED ISSUE #140: DICOM export plugin

### LINKED ISSUE #183: Import should be disabled if no master volume is selected

che85:
fixes #140 
fixes #183

### DIFF
--- a/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -3,7 +3,6 @@ import os
 import json
 import vtk
 import logging
-import getpass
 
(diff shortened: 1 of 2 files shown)


che85:
I am going to merge this for testing in the next nightly.

PR #196, opened 2017-09-14
BUG: slicer.mrmlScene.GetNodesByClass(..) causes leaks when quitting …

che85:
…Windows

* instead using slicer.util.getNodesByClass('vtkMRMLSegmentationNode')

related to https://github.com/QIICR/SlicerDevelopmentToolbox/pull/22


cpinter:
You need to call UnRegister(None) on the collection that is returned, and the object is correctly destructed on exit. It is an annoyance that this (and many other) function creates an object you're responsible for deleting, instead of you creating a collection that you can pass to the function, but it is what it is.

che85:
I absolutely agree. My workaround works pretty well now, so I will keep it like that.

PR #207, opened 2017-11-22
ENH: removed LabelmapSegmentStatisticsPlugin from custom statistics

che85:
The reason for removing that plugin from the plugins list is that Measurements were duplicated.

I implemented it the way it's done in the SegmentStatisticsTest [1] although I don't think that it should be implemented that way because it "unregisters" the SegmentStatistics plugins globally since it's set on the SegmentStatisticsLogic class directly. 

Calling SegmentStatistics module afterwards, it only those plugins, that I registered for QuantitativeReporting are displayed, but all others are gone.

[1] https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L829-L832

cc @chribaue

### DIFF
--- a/QuantitativeReporting/QRCustomizations/CustomSegmentStatistics.py
+++ b/QuantitativeReporting/QRCustomizations/CustomSegmentStatistics.py
@@ -1,8 +1,9 @@
 import logging
-
 import slicer
+
 from SlicerDevelopmentToolboxUtils.mixins import ModuleLogicMixin
 from SegmentStatistics import SegmentStatisticsLogic
+from SegmentStatisticsPlugins import ScalarVolumeSegmentStatisticsPlugin, ClosedSurfaceSegmentStatisticsPlugin
 from DICOMSegmentationPlugin import DICOMSegmentationExporter
 
 
@@ -23,6 +24,10 @@ class CustomSegmentStatisticsLogic(SegmentStatisticsLogic):
     return slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("Segmentation"))
 
   def __init__(self):
+    logging.debug("CustomSegmentStatisticsLogic: only using ScalarVolumeSegmentStatisticsPlugin and "
+                  "ClosedSurfaceSegmentStatisticsPlugin")
+    SegmentStatisticsLogic.registeredPlugins = [ScalarVolumeSegmentStatisticsPlugin,
+                                                ClosedSurfaceSegmentStatisticsPlugin]
     SegmentStatisticsLogic.__init__(self)
     self.terminologyLogic = slicer.modules.terminologies.logic()
 


che85:
@chribaue The builtin `plugins` class member must not be modifiable from the outside. You should rename it to `__plugins` and invent a mechanism for users/developers to define which plugins to use when instantiating the class.

Just add an additional parameter to `__init__(..)`, that users allow to define which plugins should be visible/used for calculating segment statistics.  If that parameter is not defined, just use the ones defined in your `__plugins` member

che85:
I merged this PR since we want that to be working for RSNA

chribaue:
Hi che85,

maybe help me better to understand the reason why you want to remove the LabelmapSegmentStatisticsPlugin and what measurements are duplicated. The way you do it right now it does not just remove the LabelmapSegmentStatisticsPlugin but also other plugins that may be registered :(, like PET statistics.

You could simply disable the one plugin that you don't like, the way it's done here:
https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L811

Or, Instead of overwriting the list of globally registered modules in your class before calling the superlass's init(), you could first call the superlass's init(), remove the plugin you don't like from self.plugins and call self.reset(). I think that should do what you want it to do. Look at the initialisier of SegmentStatisticsLogic what it does with registered modules
https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L354-L363

Hope that helps.

fedorov:
I agree with @chribaue - for this module purposes, we should only make sure that those plugins that do not use scalar volume are excluded. It will be better to take the plugin list and exclude modules we don't need.

The background is that we didn't realize, but we have duplicate measurements because Labelmap plugin calculates the same measurement as Scalar volume plugin.

che85:
Oh yeah I totally missed the other plugins that might get registered. Sorry for the confusion. Will create a new PR with the fix.

che85:
> Or, Instead of overwriting the list of globally registered modules in your class before calling the superlass's init(), you could first call the superlass's init(), remove the plugin you don't like from self.plugins and call self.reset(). I think that should do what you want it to do. Look at the initialisier of SegmentStatisticsLogic what it does with registered modules
https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L354-L363

I tried that already, but it didn't work for any reason

che85:
I think I found the reason why. `SegmentStatisticsParameterEditorDialog` is instantiating it's own `SegmentStatisticsLogic`.

che85:
`SegmentStatisticsParameterEditorDialog` is instantiating it's own `SegmentStatisticsLogic`[1]

Even if you set the `parameterNode` to the modified parameter node of the SegmentStatisticsLogic that does not include the LabelmapSegmentStatisticsPlugin, everything gets resetted by call [2] and [3] in which all default plugins will get used since a new instance of SegmentStatisticsLogic has been created. 

[1] https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L269-L275

[2] https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L274
[3] https://github.com/Slicer/Slicer/blob/08fec03c10955e813b22886697ee2cdb346898df/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L371-L378

che85:
I customized the dialog by overriding the __init__ method and that worked well [1]. One thing though that is not working right now is, opening the dialog twice 

## 1.call
image

## 2.call
image

I might take a look later again.

[1] https://github.com/che85/Reporting/commit/36ba9b176b46703f03f5a5b32a9ffab70bcfb958

fedorov:
@che85 another way to deal with this is instead of disabling plugins, go over the list of measurements, and if there is any duplication based on the combination of code+designator and modifier code+designator, then drop the duplicate. Do you see what I mean?

lassoan:
I did not follow the all the discussion above, so I just comment on this topic in general.

Slicer does not create a logic class for scripted modules on startup. Therefore, you cannot store the global SegmentStatistics plugin list in the SegmentStatistics module logic (you can of course have a list of plugins in the module logic that is initialized by default from the global SegmentStatistics plugin list). 

In scripted modules, we usually store global information like these in the `slicer` namespace. For example, for self-tests are registered like this:

```
try:
  slicer.selfTests
except AttributeError:
  slicer.selfTests = {}
slicer.selfTests['RSNA2012ProstateDemo'] = self.runTest
```

The same approach could work for registering SegmentStatistics plugins (but probably using a simple list instead of a map).

Maybe you could add a static function in SegmentStatistics.py that does this, so that you don't need to repeat these couple of lines at each place you register a plugin.

chribaue:
I agree with @fedorov, if multiple plugins produce measurements with the same DICOM codes and this causes troubles during exporting right now, simply export one of the results. The measurement values themselves should be identical.

I don't think it's a good idea that QuantiativeReporting restricts what Plugins I can use.
If you start disabling/removing or only enabling certain plugins, this certainly will confuse users or make it impossible for them to use certain plugins.

For example, I certainly want to use our PETVolumeSegmentStatisticsPlugin (which will calculates a 
"volume" measure) in the QuantiativeReporting module and I still want to have the option to use the LabelmapSegmentStatisticsPlugin as well.

fedorov:
@chribaue I agree with your point, and I also apologize the PR was merged very quickly and didn't leave time for the discussion. We discovered this issue very late in time relative to RSNA, and wanted to fix it quickly. I discussed with @che85 and we both agreed filtering of duplicated measurements is the right way going forward. That approach will be implemented after RSAN - we just didn't want to push it out at the risk of breaking something else.

PR #208, opened 2017-11-23
ENH: removing unneeded plugins only from instantiated SegmentStatisticsLogic

che85:
for reference to prior discussion https://github.com/QIICR/QuantitativeReporting/pull/207


che85:
@fedorov It's working with the disabling now. I just had to keep the dialog as a class member. Qt ... I am going to merge this, so that it will be available in tomorrow's build

Issue #203, opened 2017-11-20
QuantitativeReporting breaks DICOM widget in Slicer 4.8


lassoan:
When installing QuantitativeReporting extension for Slicer 4.8 (Windows) and switching to DICOM module, the browser does not appear.

```
[DEBUG][Qt] 20.11.2017 10:54:10 [] (unknown:0) - Session start time .......: 2017-11-20 10:54:10
[DEBUG][Qt] 20.11.2017 10:54:10 [] (unknown:0) - Slicer version ...........: 4.8.0 (revision 26489) win-amd64 - installed
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - Operating system .........: Windows /  Professional /  (Build 9200) - 64-bit
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - Memory ...................: 16302 MB physical, 48302 MB virtual
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - CPU ......................: GenuineIntel , 8 cores, 16 logical processors
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - Developer mode enabled ...: yes
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - Prefer executable CLI ....: yes
[DEBUG][Qt] 20.11.2017 10:54:11 [] (unknown:0) - Additional module paths ..: C:/Users/msliv/AppData/Roaming/NA-MIC/Extensions-26489/QuantitativeReporting/lib/Slicer-4.8/qt-scripted-modules, C:/Users/msliv/AppData/Roaming/NA-MIC/Extensions-26489/SlicerDevelopmentToolbox/lib/Slicer-4.8/qt-scripted-modules, C:/Users/msliv/AppData/Roaming/NA-MIC/Extensions-26489/DCMQI/lib/Slicer-4.8/cli-modules
[DEBUG][Python] 20.11.2017 10:54:12 [Python] (C:\Program Files\Slicer 4.8.0\lib\Python\Lib\site-packages\git\cmd.py:719) - Popen(%s, cwd=%s, universal_newlines=%s, shell=%s)
[DEBUG][Python] 20.11.2017 10:54:12 [Python] (C:\Program Files\Slicer 4.8.0\lib\Python\Lib\site-packages\git\cmd.py:719) - Popen(%s, cwd=%s, universal_newlines=%s, shell=%s)
[...]

PR #212, opened 2017-12-04
BUG: specified resource files explicitly instead of glob (issue #203)

che85:
* removed unused(outdated) files

work on #203


che85:
Made a mistake. will create a new one.

che85:
@fedorov Let me know if you have any concerns regarding the outdated files that I removed

fedorov:
No concerns

PR #226, opened 2018-03-26
Algorithm tracking

chribaue:
ENH: functionality to specify DICOM algorithm name and type and export to DICOM SEG

### DIFF
--- a/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -205,6 +205,10 @@ class DICOMSegmentationPluginClass(DICOMPluginBase):
           labelNode.SetAttribute("ColorR", str(rgb[0]))
           labelNode.SetAttribute("ColorG", str(rgb[1]))
           labelNode.SetAttribute("ColorB", str(rgb[2]))
+          if "SegmentAlgorithmType" in segment:
+            labelNode.SetAttribute("SegmentAlgorithmType", segment["SegmentAlgorithmType"])
+          if "SegmentAlgorithmName" in segment:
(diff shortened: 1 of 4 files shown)


che85:
@chribaue Just out of curiosity, what if the user selects a tool, uses it to edit a segment and then decides to undo the changes and select another tool?

DICOMPlugins/DICOMSegmentationPlugin.py
```diff
@@ -205,6 +205,10 @@ def load(self,loadable):
           labelNode.SetAttribute("ColorR", str(rgb[0]))
           labelNode.SetAttribute("ColorG", str(rgb[1]))
           labelNode.SetAttribute("ColorB", str(rgb[2]))
+          if "SegmentAlgorithmType" in segment:
+            labelNode.SetAttribute("SegmentAlgorithmType", segment["SegmentAlgorithmType"])
+          if "SegmentAlgorithmName" in segment:
+            labelNode.SetAttribute("SegmentAlgorithmName", segment["SegmentAlgorithmName"])
```
fedorov: Is there a good reason to not use `DICOM.` prefix here, but then use it later while assigning attributes to the `segmentLabelNode`? Shouldn't it be consistent?

fedorov:
@chribaue this looks good to me, just one minor comment.

DICOMPlugins/DICOMSegmentationPlugin.py
chribaue: `labelNode` is just temporary data, but I'll add the `DICOM.` prefix for readability

fedorov: Great, thanks - I have no other concerns.

PR #228, opened 2018-04-23
WIP: Meta data form

### LINKED ISSUE #201: specifying meta-data for DICOM Segs in Quantitative Reporting module
The exported DICOM Segs currently store proper information what the
segments encode and from which dataset they were derived. However, there
are still items that need to be done to properly encode who produced the
segmentation for what purpose and with which tool/algorithm. Browsing
through your code, I can see that during development of dcmqi and the
Quantitative Reporting module you already planned ahead for supporting
these items, but so far, several meta-data properties are only set to
generic default values and there's no way for the user to specify them
through the Slicer GUI.

https://qiicr.gitbooks.io/dcmqi-guide/user_guide/itkimage2segimage.html
neatly lists all these properties that should be properly specified:

general content information:
  * "ContentCreatorName"
  * "ClinicalTrialSeriesID"
  * "ClinicalTrialTimePointID"
  * "ClinicalTrialCoordinatingCenterName"
  * "SeriesDescription"
  * "SeriesNumber"
  * "InstanceNumber"

for each segment:
  * "SegmentAlgorithmType"
  * "SegmentAlgorithmName"

I guess for the "general information" a simple dialog box and a
mechanism that stores the specified information between Slicer sessions
will suffice.

Regarding information about algorithm type and name, I would say this
information should be provided by the producing algorithm and I would
propose that algorithms that want to provide this information should add
proper information to the Segment's tags, i.e.:
  * segment.SetTag('DICOM.SegmentAlgorithmName', "PET Tumor Segmentation
v1.0")
  * segment.SetTag('DICOM.SegmentAlgorithmType', "SEMIAUTOMATIC")

During exporting DICOM Segs this information would be used, if present, and
default values would be used otherwise.

che85:
fixes #201

- [x] general information from the meta-data dialog does not get stored (e.g. ClinicalTrialCoordinatingCenterName) between sessions. This would be very useful for the user.
- [x] add instruction on the top what this form is about "Please enter DICOM SEG meta-data...."
- [x] handle empty input
- [x] check if user info was set and notify user if not

- [x] ClinicalTrialSeriesID doesn't get saved to the DICOM SEG (fixed in https://github.com/QIICR/dcmqi/pull/344)
- [x] add SeriesNumber and InstanceNumber to the dialog box
- [x] move SeriesDescription from the bottom of the list to the second position (after ContentCreatorName)
- [x] restore JSON defaults (https://github.com/QIICR/SlicerDevelopmentToolbox/commit/8f36a9b90b039d29bef96dca49e2d938c3db6e26)


che85:
@chribaue Maybe you can go ahead, test and give feedback? I know it's not perfect yet, but working. 

TODO: I still need to take care of situations like empty fields

chribaue:
@chea85 Great! I'll take a look

che85:
Basically, when saving/completing report the form will be displayed. Please make sure to use the most recent nightly Slicer which includes the UserInformation.

P.S. I should also add a check if user information were set

che85:
@chribaue You can test again. One item of the list is not quite ready yet because I need to think about how to cache the information across sessions.  

Please download the nightly Slicer tomorrow to make sure that you have the most recent version of dcmqi.

chribaue:
@che85 works like a charm now for writing proper meta-data to the DICOM SEGs. Great work!

I'm not sure if its related, but the tid1500 writer crashed when exporting the SR. I updated my dcmqi.

che85:
> I'm not sure if its related, but the tid1500 writer crashed when exporting the SR. I updated my dcmqi.

What exactly do you mean with that? Can you give me some instructions for reproducing it?

chribaue:
I loaded a PET H&N scan, segmented 2 lesions and clicked "complete report" and "Ok". Below is the command line output I get. I just tested again, I doesn't happen when I segment only 1 lesion.

```
/home/christian/software/dcmqi-build/dcmqi-build/lib/Slicer-4.9/cli-modules/tid1500writer --inputMetadata /tmp/Slicer-christian/__SlicerTemp__/2018-05-16_103453/sr_meta.json --outputDICOM /tmp/Slicer-christian/__SlicerTemp__/2018-05-16_103453/sr.dcm --inputCompositeContextDirectory /tmp/Slicer-christian/__SlicerTemp__/2018-05-16_103453 --inputImageLibraryDirectory ./ctkDICOM-Database/TCIALocal/1/1.3.6.1.4.1.14519.5.2.1.2744.7002.135516888288831261489636271190/images 

DICOM Measurements reporting standard output:

dcmqi repository URL: https://github.com/QIICR/dcmqi revision: 4bd1996 tag: latest-2-g4bd1996
Total measurement groups: 2

DICOM Measurements reporting standard error:

Condition failed: Illegal parameter in /home/christian/software/dcmqi/apps/sr/tid1500writer.cxx:210
terminate called after throwing an instance of 'int'



DICOM Measurements reporting terminated with an unknown exception.



Traceback (most recent call last):
  File "/home/christian/software/Reporting-build/lib/Slicer-4.9/qt-scripted-modules/QuantitativeReporting.py", line 616, in onCompleteReportButtonClicked
    success, err = self.saveReport(completed=True)
  File "/home/christian/software/Reporting-build/lib/Slicer-4.9/qt-scripted-modules/QuantitativeReporting.py", line 631, in saveReport
    dcmSRPath = self.createDICOMSR(dcmSegPath, completed)
  File "/home/christian/software/Reporting-build/lib/Slicer-4.9/qt-scripted-modules/QuantitativeReporting.py", line 709, in createDICOMSR
    raise Exception("tid1500writer CLI did not complete cleanly")
Exception: tid1500writer CLI did not complete cleanly
```

che85:
Can you send the Slicer error log?

chribaue:
I think there might be "nan" values in the measurements table. Would this cause issues?

fedorov:
Ah, yes - it certainly would ... and there is an unresolved PR for this: https://github.com/QIICR/dcmqi/pull/290. I don't recall why that PR got stalled ...

chribaue:
I just tested again. Without "nan" values it works also for 2 lesions. Good that we figured this out and there's a fix for this issue already. I guess we don't need the error log anymore.

fedorov:
@chribaue I fixed and merged the PR referenced above. The converter should now automatically skip all invalid measurements.

chribaue:
@che85 Did you have time for:
> I need to think about how to cache the information across sessions.

fedorov:
@chribaue Christian left BWH last week, I am not sure if he will be spending any more time on this... He is now traveling, and will be starting his new job in September.

chribaue:
@fedorov I'll take a look how to add the remaining item..

che85:
@chribaue I didn't get to that because I switched to another computer in the meanwhile. If you want, I can still take a look.

che85:
@chribaue How would you like it to look like? As you know it currently looks as follows. 

screen shot 2018-30-5 at 12-05-06

Two thoughts:
1. We could temporarily save all entered information every time the dialog gets used and approved.
2. We could add checkboxes as I proposed earlier. (see https://github.com/QIICR/QuantitativeReporting/issues/201#issuecomment-357144637)

For both, I think I would use the qSettings to store the information across sessions (still available after restarting)

P.S. we could also add a button for restoring the defaults.

fedorov:
I agree 1 is better. I also think that in most cases users would not care about the defaults.

Note that `ContentCreatorName` is PN VR, so it would need to be `Herz^Christian` for the example above. See http://dicom.nema.org/Dicom/2013/output/chtml/part05/sect_6.2.html#sect_6.2.1.1 for the examples.

che85:
@chribaue that should do the trick. Please test and let me know. 

P.S. make sure to pull the new commit from https://github.com/QIICR/SlicerDevelopmentToolbox/commit/8f36a9b90b039d29bef96dca49e2d938c3db6e26

chribaue:
@che85 saving between sessions work now :+1: All items on the list have been completed :tada: 

Regarding content creator name and @fedorov's comment above, the software turns the entered `ContentCreatorname` "Christian Herz" into `Herz^Christian` as it should, but how does the software know how in what format the user entered the name? What if I enter e.g. "Herz, Christian"? Should we provide an example for users in what format to enter the name?

fedorov:
Maybe separate fields for first and last name?

lassoan:
In the application settings, we need to use the DICOM standard to encode first, last, middle names, prefixes, and suffixes (http://dicom.nema.org/dicom/2013/output/chtml/part05/sect_6.2.html#sect_6.2.1.1), otherwise storage could become very complicated.

To make the GUI more user-friendly, we could either decode/encode to a human-readable string (DCMTK already has DICOM to human-readable converter (https://support.dcmtk.org/docs/classDcmPersonName.html#aa7635519dff43d43e1e0e10252843f4c), or have multiple fields on the GUI only (that would be read/written as a single DICOM-formatted string in the settings).

chribaue:
I guess, adding to `vtkPersonInformation` fields `lastName`, `firstName`, `middleName`, `namePrefix`, `nameSuffix`, and updating `qSlicerSettingsUserInformationPanel` and our DICOM Seg meta-data form should be strait forward.

What I'm not sure about is what to do with the current `GetName`/`SetName` in `vtkPersonInformation`. Should they remain or get removed? Or should we keep a `GetName` that returns a properly formatted representation of the user entered details?

@fedorov I think we should still merge @che85's PR now already as it is  and make upgrade as the Slicer application settings become more DICOM conform. What do you think?

fedorov:
> I guess, adding to vtkPersonInformation fields lastName, firstName, middleName, namePrefix, nameSuffix, and updating qSlicerSettingsUserInformationPanel and our DICOM Seg meta-data form should be strait forward.

It is definitely straightforward, but I this will pollute UI (and API) without clear utility. If we start to expose all possible options that are available in DICOM examples, we will very very quickly get the user overwhelmed. In this particular case, suffix/prefix etc are derived from the examples, and are not even formalized as separate attributes. 

Initially, I thought about adding separate fields for the first and last name, but after thinking about this more, in the spirit of simplifying the UI/API, I suggest we keep a single field for the name, and if the input has a space, we should replace it with '^' on "Ok" and notify the user, or have a validator that does not allow spaces. In the use cases I can think of, identification of the reader most likely will be a single word ID unique within the project, so in most cases this will not be a concern. I don't think anyone should rely on this attribute to uniquely identify the reader within the universe.

> I think we should still merge @che85's PR now already as it is and make upgrade as the Slicer application settings become more DICOM conform. 

Yes, this is fine with me.

chribaue:
> we should replace it with '^' 

That's @che85's code already does.

> In the use cases I can think of, identification of the reader most likely will be a single word ID unique within the project,

I agree. People doing annotations for us would not enter their real name but a project related user ID

From our end, the current DICOM Seg meta-data form already fulfills all the requirements we need.

fedorov:
Sounds good, I suggest we keep it as is until there is a specific need to revisit.

lassoan:
Storage should be kept as DICOM (no need to reinvent a new scheme) and vtkPersonInformation should allow reading/writing this raw DICOM format directly. Everything else is for developer and user convenience and can be gradually improved.

For user convenience, it could work like Outlook, which has automatic parsing from single string:

image

but if you click on the Full name button (or the entered string is ambiguous) then you get a detailed GUI:

image

For developer's convenience, we could add GetDisplayableName/SetFromDisplayableName; and Get/Set functions for FirstName, LastName, Prefix, Suffix, etc.

PR #229, opened 2018-05-07
ENH: added delayed auto update of measurements to avoid unnecessary c…

chribaue:
…alculations

### DIFF
--- a/QuantitativeReporting/QuantitativeReporting.py
+++ b/QuantitativeReporting/QuantitativeReporting.py
@@ -61,6 +61,16 @@ class QuantitativeReportingWidget(ModuleWidgetMixin, ScriptedLoadableModuleWidge
     self.slicerTempDir = slicer.util.tempDirectory()
     slicer.mrmlScene.AddObserver(slicer.mrmlScene.EndCloseEvent, self.onSceneClosed)
 
+    # Wait this much after the last modified event before starting updating measurements
+    autoUpdateDelaySec = 0.5
+    self.delayedAutoUpdateTimer = qt.QTimer()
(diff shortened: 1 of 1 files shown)


QuantitativeReporting/QuantitativeReporting.py
```diff
@@ -61,6 +61,16 @@ def __init__(self, parent=None):
     self.slicerTempDir = slicer.util.tempDirectory()
     slicer.mrmlScene.AddObserver(slicer.mrmlScene.EndCloseEvent, self.onSceneClosed)
 
+    # Wait this much after the last modified event before starting updating measurements
+    autoUpdateDelaySec = 0.5
+    self.delayedAutoUpdateTimer = qt.QTimer()
+    self.delayedAutoUpdateTimer.setSingleShot(True)
+    self.delayedAutoUpdateTimer.interval = autoUpdateDelaySec * 1000
+    self.delayedAutoUpdateTimer.connect('timeout()', self.updateMeasurementsTable)
+
```
che85: You can just use the following line. The functionality of creating a timer is provided by the SlicerDevelopmentToolbox (https://github.com/QIICR/SlicerDevelopmentToolbox/blob/29ba8592cec9b651bbcd5e1693cd1b8c075d30c1/SlicerDevelopmentToolboxUtils/mixins.py#L120-L125)
```
self.delayedAutoUpdateTimer = self.createTimer(500, self.updateMeasurementsTable, singleShot=True)

```

chribaue: @che85 makes sense in the QR module since it already depends on the DevelopmentToolboxUtils

PR #235, opened 2018-08-31
ENH: Adds length measurement import

pieper:
This is a first pass at reading imaging measurement reports.
Currently length measurements are mapped to annotation rulers.

This has been tested using the measurements converted
by @dclunie as part of the CrowdsCureCancer project.


fedorov:
@pieper I am thinking we should wait until the structure of the CrowdsCureCancer project SRs is finalized, and we have a public sample dataset that can be used for testing this feature. What do you think?

pieper:
Right, there's no particular rush.  @Erik Ziegler <erik@radicalimaging.com> and
I are working on the javascript version of this and we'll have more worked
out examples as a result of that effort.

fedorov:
Sounds good - once David confirms the final structure, and the sample datasets are established, please update this PR so I know what to test it against! Thank you!

fedorov:
@malaterre FYI this is the PR I mentioned in the email exchange earlier today so you can monitor it directly.

malaterre:
@nroduit Just FYI CrowdsCureCancer folks are working on DICOM SR with measurement (non-image based).

pieper:
@fedorov shall we find some time to talk about merging this?  It's not a blocking issue before merging but we should discuss how we want to handle variants of report and measurement types.

fedorov:
Steve, I've been working to migrate/reorganize DICOM4QI content, and want to finish that first, but I will get to testing this right after, now that I have the confirmed sample dataset from David, which I will also add to DICOM4QI.

fedorov:
@pieper as I was combing through email threads, it looks like I still do not have a sample SR with the matching image series. Can you share one with me? David only sent the sample SR.

fedorov:
@pieper I searched this page for "Crowds", but didn't find the matching collection: https://www.cancerimagingarchive.net/. Can you point me to the location of data?

fedorov:
ah, google found it: https://wiki.cancerimagingarchive.net/display/DOI/Crowds+Cure+Cancer%3A+Data+collected+at+the+RSNA+2017+annual+meeting

fedorov:
For the record, it is quite awkward to load and trace back stuff in that collection, since the only option is to download the whole dataset of images, which I really don't need.

So to locate the matching image, I opened the SR provided by David, looked up patient ID, from Study figured out it is a Kidney annotation, then searched TCIA main page for "Renal", which led me to TCIA-KIRC collection, which I then searched for subject TCGA-BP-4343. The study for that subject contained 6 series, and I located the matching one by UID ... what a pain. Why would they not put those SRs into TCIA DICOM database proper?

pieper:
> Why would they not put those SRs into TCIA DICOM database proper?

Good question for Justin?

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -332,6 +360,80 @@ def enumerateDuplicateNames(self, items):
       items[idx]["name"] = item
     return items
 
+  def isConcept(self, item, coding):
+    code = item.ConceptNameCodeSequence[0]
+    return code.CodingSchemeDesignator == self.codings[coding]["scheme"] and code.CodeValue == self.codings[coding]["value"]
+
+  def loadAdditionalMeasurements(self, srUID):
```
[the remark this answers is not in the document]
fedorov: Would it be possible to refactor this function a bit so that it is more intuitive how to add support for other types of planar annotations?

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -332,6 +360,80 @@ def enumerateDuplicateNames(self, items):
...
+          if self.isConcept(contentItem, "measurementGroup"):
+            measurement = {}
+            for measurementItem in contentItem.ContentSequence:
+              if self.isConcept(measurementItem, "trackingIdentifier"):
+                measurement['trackingIdentifier'] = measurementItem.TextValue
+              if self.isConcept(measurementItem, "trackingUniqueIdentifier"):
+                measurement['trackingUniqueIdentifier'] = measurementItem.UID
+              if self.isConcept(measurementItem, "findingSite"):
+                measurement['findingSite'] = measurementItem.ConceptCodeSequence[0].CodeMeaning
+              if self.isConcept(measurementItem, "length"):
```
fedorov: I think it is more appropriate and robust to only consider NUM type content items while looking for measurements.

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -332,6 +360,80 @@ def enumerateDuplicateNames(self, items):
...
+                for lengthItem in measurementItem.ContentSequence:
+                  measurement['polyline'] = lengthItem.GraphicData
+                  for selectionItem in lengthItem.ContentSequence:
+                    if selectionItem.RelationshipType == "SELECTED FROM":
```
fedorov: I am a bit confused as to how this parses the content - should it be checking for the presence of SCOORD in the first place?

```text
  <contains CONTAINER:(,,"Imaging Measurements")=SEPARATE>
    <contains CONTAINER:(,,"Measurement Group")=SEPARATE>
      <has obs context TEXT:(,,"Tracking Identifier")="5b6eb4301d3175942d29985a3d0...">
      <has obs context UIDREF:(,,"Tracking Unique Identifier")="1.3.6.1.4.1.5962.1.1.0.0.0.1535644357.22655.1">
      <has concept mod CODE:(,,"Finding Site")=(T-71000,SRT,"Kidney")>
      <contains NUM:(,,"Length")="66.43856134" (mm,UCUM,"mm")>
        <inferred from SCOORD:=(POLYLINE,172.835/270.0641,...)>
          <selected from IMAGE:=(CT image,)>
```

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -332,6 +360,80 @@ def enumerateDuplicateNames(self, items):
...
+      rulerNode.Initialize(slicer.mrmlScene)
+      colorIndex = 1 + len(slicer.util.getNodesByClass('vtkMRMLAnnotationRulerNode'))
+      colorNode = slicer.util.getNode("GenericAnatomyColors")
+      color = [0,]*4
+      colorNode.GetColor(colorIndex, color)
+      rulerNode.GetDisplayNode().SetColor(*color[:3])
+      rulerNode.GetAnnotationTextDisplayNode().SetColor(*color[:3])
+      rulerNode.GetAnnotationPointDisplayNode().SetColor(*color[:3])
+      rulerNode.GetAnnotationPointDisplayNode().SetGlyphScale(2)
+      rulerNode.GetAnnotationLineDisplayNode().SetLabelPosition(random.random())
```
fedorov: I remember in the past we had a pointer from fiducials to the SOPInstanceUIDs when associated with DICOM. Can something like this be added here as well?

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -127,9 +141,20 @@ def sortReportsByDateTime(self, uids):
...
-    seriesTime = dataset.SeriesTime
-    return datetime.datetime.strptime(seriesDate+seriesTime, '%Y%m%d%H%M%S')
+    if hasattr(dataset, 'SeriesDate') and hasattr(dataset, "SeriesTime"):
+      date = dataset.SeriesDate
+      time = dataset.SeriesTime
+    elif hasattr(dataset, 'StudyDate') and hasattr(dataset, "StudyTime"):
+      date = dataset.StudyDate
+      time = dataset.StudyDate
+    else:
+      date = "19630524"
```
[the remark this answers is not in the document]
pieper: I changed the default to be empty strings when the value is unknown, which is what the [standard dictates](https://dicom.innolitics.com/ciods/rt-dose/general-study/00080020).  This should be fine since the string is only used for sorting.

DICOMPlugins/DICOMTID1500Plugin.py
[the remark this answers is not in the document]
pieper: Let's do that when we have some actual documents to test against.

DICOMPlugins/DICOMTID1500Plugin.py
pieper: Let's generate (or collect) some examples of the types of documents we want to support and then we can generalize the code.

DICOMPlugins/DICOMTID1500Plugin.py
pieper: Yes, that would make sense at some point.  We should visit that when we add export because I really think we want round-trip support.  This really needs to be migrated to the Markups anyway.

DICOMPlugins/DICOMTID1500Plugin.py
```diff
@@ -332,6 +360,80 @@ def enumerateDuplicateNames(self, items):
...
+                        if hasattr(reference, "ReferencedFrameNumber") and reference.ReferencedFrameNumber != "1":
+                          print('Error - only single frame references supported')
+            measurements.append(measurement)
+
+    for measurement in contents['measurements']:
+      rulerNode = slicer.vtkMRMLAnnotationRulerNode()
+      rulerNode.SetLocked(True)
+      rulerNode.SetName(contents['personObserver'])
+
+      referenceFilePath = slicer.dicomDatabase.fileForInstance(measurement['referencedSOPInstanceUID'])
```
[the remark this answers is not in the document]
pieper: Created issue for this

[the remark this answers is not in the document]
pieper: https://github.com/QIICR/QuantitativeReporting/issues/248

DICOMPlugins/DICOMTID1500Plugin.py
pieper: created issue https://github.com/QIICR/QuantitativeReporting/issues/249

PR #239, opened 2019-07-08
ENH: updates to make module work with python 3

fedorov:
depends on https://github.com/QIICR/SlicerDevelopmentToolbox/pull/36

### DIFF
--- a/DICOMPlugins/DICOMParametricMapPlugin.py
+++ b/DICOMPlugins/DICOMParametricMapPlugin.py
@@ -105,7 +105,7 @@ class DICOMParametricMapPluginClass(DICOMPluginBase):
       self.cleanup()
       return False
 
-    (_,pmNode) = slicer.util.loadVolume(os.path.join(self.tempDir,"pmap.nrrd"), returnNode=True)
+    pmNode = slicer.util.loadVolume(os.path.join(self.tempDir,"pmap.nrrd"))
 
     # load the metadata JSON to retrieve volume semantics (quantity stored and units)
     with open(os.path.join(self.tempDir,"meta.json")) as metafile:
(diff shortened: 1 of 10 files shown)


fedorov:
@che85 do you have any idea how to fix this error:

image

che85:
What versions of Slicer and the repositories are you using? That way, it might be easier to reproduce on my side

fedorov:
Today's nightly with this branch, and SDR branch mentioned above.

fedorov:
Yes, I agree - those were added automatically by that modernizer package I mentioned in the SDT PR https://github.com/mitsuhiko/python-modernize. I thought it is probably more efficient to use that script instead of manually identifying and fixing one issue at a time, which I did to fix mpReview...

che85:
@fedorov I added some code. The error was caused by ScriptedLoadableModule trying to get the modulePath for looking for resources. A new feature was added that automatically provides `resourcesPath` for any subclassed ScriptedLoadableModule

fedorov:
Thanks Christian - working better now - next issue is this - happens if I examine a DICOM SEG instance:

image

fedorov:
The above seemed to be fixed by the patch below using `python-modernizer`:

```
--- ./SlicerDevelopmentToolboxUtils/decorators.py	(original)
+++ ./SlicerDevelopmentToolboxUtils/decorators.py	(refactored)
@@ -1,8 +1,11 @@
+from __future__ import absolute_import
+from __future__ import print_function
 from functools import wraps
 import itertools
 import logging
 import inspect
 import slicer
+from six.moves import map


 class logmethod(object):
@@ -226,7 +229,7 @@

   def getStringifiedClasses(t):
     if type(t) is list:
-      return map(getStringifiedClasses, t)
+      return list(map(getStringifiedClasses, t))
     else:
       return t.__name__ if not type(t) is str else t

@@ -250,7 +253,7 @@
     startTime = time.time()
     x = func(*args, **kwargs)
     duration = time.time() - startTime
-    print("{} ran in: {0} seconds".format(func.__name__, duration))
+    print(("{} ran in: {0} seconds".format(func.__name__, duration)))
     return x

   return _new_function
```

PR #245, opened 2020-02-28
COMP: updated code and self test

chribaue:
* fixed self tests related to changes in Slicer's API
* in python3 exceptions no longer have a message attribute

### DIFF
--- a/Testing/QuantitativeReportingTests.py
+++ b/Testing/QuantitativeReportingTests.py
@@ -42,9 +42,10 @@ class QuantitativeReportingTests:
       slicer.selfTests = {}
     slicer.selfTests['QuantitativeReporting'] = self.runTest
 
-  def runTest(self):
+  def runTest(self, msec=100, **kwargs):
     tester = QuantitativeReportingTest()
-    tester.runTest()
+    tester.messageDelay = msec
+    tester.runTest(**kwargs)
 
 
 class QuantitativeReportingTestsWidget(ScriptedLoadableModuleWidget):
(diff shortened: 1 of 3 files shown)


Testing/QuantitativeReportingTests.py
```diff
@@ -158,7 +159,8 @@ def loadTestData():
       cacheAutoLoadReferences = qt.QSettings().value('DICOM/automaticallyLoadReferences')
       qt.QSettings().setValue(settings, qt.QMessageBox.Yes)
 
-      qrWidget.loadSeries(self.data['sr']['uid'])
+      qrWidget.loadSeries(self.data['sr']['uid']) # note: this no longer loads referenced data recursively; this functionality only works whithin the DICOMbrowser GUI
+      qrWidget.loadSeries(self.data['volume']['uid']) # load volume manually for test
```
fedorov: Should SEG be explicitly loaded then as well?

chribaue: That SEG is loaded automatically.

The SR references the SEG, which gets loaded but the volume that is in turn referenced by the SEG doesn't get loaded. This recursive loading seems to be intended only for the UI, and can't get accessed from outside the UI anymore.
https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOMLib/DICOMBrowser.py#L481-L518.

fedorov: @lassoan is this change in behavior intentional?

lassoan: GUI and logic was intertwined in the old DICOM browser. I separated them to allow scripts to import/load/export DICOM without relying on any GUI classes.

If you find that some more logic could be pushed down from GUI classes to logic then let me know - the best would be if you could send a pull request with the proposed changes.

PR #252, opened 2020-09-24
Fix setting of segment name in when loading segmentation from DICOM

### LINKED ISSUE #251: Segment name is not stored in exported segmentation object
I’m helping Dr Simpson with developing script for converting mhd files to DICOM segmentation objects for uploading to TCIA. The script uses Slicer QuantitativeReporting extension and it works nicely, it creates valid DICOM files that can be imported and loaded successfully.

However, I noticed that after a roundtrip DICOM export/import, the segment names are changed. For example, from the original “Liver”, “Mass”, “Vein” to “Anatomical structure", “Morphologically altered structure", “Tissue".

I’ve checked the code and it seems that “SegmentLabel” DICOM tag is set from segmentation property type, and “SegmentDescription” DICOM tag is set from segmentation property category. Segment name that is set in Slicer is not stored anywhere.

I would expect that segment name (that the user can specify in Slicer in addition to the standard terminology) would be stored in either SegmentLabel or SegmentDescription field. Property category and type are stored in terminology fields anyway, so we would not lose anything by not storing them in SegmentLabel and SegmentDescription.

Could we change the DICOM segmentation exporter in QuantitativeReporting extension to save segment name in SegmentLabel or SegmentDescription DICOM fields? I can implement is and send a pull request, just want to make sure that this is the right thing to do.

lassoan:
Problems fixed:
- Segment description or code meaning was used instead of segment label.
- When the same label was used multiple times, _1, _2, ... suffixes were added (because segment name was passed via node name).

Fixes #251

### DIFF
--- a/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -180,17 +180,18 @@ class DICOMSegmentationPluginClass(DICOMPluginBase):
           # files and labelmap nodes
           # if not hasattr(slicer.modules, 'segmentations'):
 
-          # load the segmentation volume file and name it for the reference series and segment color
           labelFileName = os.path.join(self.tempDir, str(segmentId) + ".nrrd")
-          if segment["SegmentDescription"] is None:
-            segmentName = seriesName + "-" + typeCodeMeaning + "-label"
-          else:
-            segmentName = segment["SegmentDescription"]
-
-          labelNode = slicer.util.loadLabelVolume(labelFileName,properties={'name': segmentName})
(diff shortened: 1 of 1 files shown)


fedorov:
@lassoan in #251 you said "Could we change the DICOM segmentation exporter in QuantitativeReporting extension to save segment name in SegmentLabel or SegmentDescription DICOM fields?", but I do not see any changes related to that export step. So it seems that this PR would not fix the issue you identified, or am I missing something?

lassoan:
Yes, sorry, I've pushed a fix just now that includes the export step change as well.

fedorov (approved):

lassoan:
Thanks for the review.

PR #254, opened 2020-10-27
Fix tid1500 plugin

lassoan:
TID1500 and Parametric mapping plugins are updated to work with latest Slicer-4.11 version.

### DIFF
--- a/DICOMPlugins/DICOMParametricMapPlugin.py
+++ b/DICOMPlugins/DICOMParametricMapPlugin.py
@@ -71,7 +71,7 @@ class DICOMParametricMapPluginClass(DICOMPluginBase):
     logging.debug('DICOM PM load()')
     try:
       uid = loadable.uid
-      logging.debug('in load(): uid = ', uid)
+      logging.debug('in load(): uid = %s', uid)
     except AttributeError:
       return False
 
@@ -83,7 +83,7 @@ class DICOMParametricMapPluginClass(DICOMPluginBase):
 
     pmFileName = slicer.dicomDatabase.fileForInstance(uid)
     if pmFileName is None:
-      logging.debug('Failed to get the filename from the DICOM database for ', uid)
+      logging.debug('Failed to get the filename from the DICOM database for %s', uid)
       self.cleanup()
       return False
 
--- a/DICOMPlugins/DICOMTID1500Plugin.py
+++ b/DICOMPlugins/DICOMTID1500Plugin.py
@@ -30,7 +30,7 @@ class DICOMTID1500PluginClass(DICOMPluginBase, ModuleLogicMixin):
     self.codings = {
       "imagingMeasurementReport": { "scheme": "DCM", "value": "126000" },
       "personObserver": { "scheme": "DCM", "value": "121008" },
-      "imagingMeasuremnts": { "scheme": "DCM", "value": "126010" },
+      "imagingMeasurements": { "scheme": "DCM", "value": "126010" },
       "measurementGroup": { "scheme": "DCM", "value": "125007" },
       "trackingIdentifier": { "scheme": "DCM", "value": "112039" },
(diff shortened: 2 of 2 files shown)


DICOMPlugins/DICOMParametricMapPlugin.py
```diff
@@ -71,7 +71,7 @@ def load(self,loadable):
     logging.debug('DICOM PM load()')
     try:
       uid = loadable.uid
-      logging.debug('in load(): uid = ', uid)
+      logging.debug('in load(): uid = %s', uid)
```
pieper, line 74: This isn't the correct syntax; would need to use % to format the string.  Previously would just print the value, which was the expected behavior.  Modern python would prefer to use the format string `f"uid = {uid}"`

DICOMPlugins/DICOMParametricMapPlugin.py
```diff
@@ -83,7 +83,7 @@ def load(self,loadable):
 
     pmFileName = slicer.dicomDatabase.fileForInstance(uid)
     if pmFileName is None:
-      logging.debug('Failed to get the filename from the DICOM database for ', uid)
+      logging.debug('Failed to get the filename from the DICOM database for %s', uid)
```
pieper, line 86: Same as above, should use `f""`

pieper (commented):
This looks great, thanks for updating. 👍   A few python coding notes inline.

There hasn't traditionally been a diverse set of data to test with, but this plugin had been working with the SRs from the Crowds Cure Cancer dataset and it would be good to confirm at some point that this still works.

@fedorov is the CCC data on the list to include in IDC?

https://wiki.cancerimagingarchive.net/display/DOI/Crowds+Cure+Cancer%3A+Data+collected+at+the+RSNA+2017+annual+meeting

https://wiki.cancerimagingarchive.net/display/DOI/Crowds+Cure+Cancer%3A+Data+collected+at+the+RSNA+2018+annual+meeting

fedorov:
Sorry, I should have not merged before Steve's review. I will fix those two lines.

> is the CCC data on the list to include in IDC?

There are many items on the list, but we need to finalize versioning plan before we can include any additional collections.

fedorov:
> is the CCC data on the list to include in IDC?

Looking at the links you shared, the SR files are in zip files, they are not in TCIA DICOM collections. Why is that?

Do we know if those annotations work with the current OHIF / IDC Viewer? We will definitely want to know this before we include those instances in IDC.

pieper:
I haven't tried the CCC annotations with OHIF.  I'm sure it can be made to work, but they may be slightly different as @lassoan pointed out.  I'm not sure why Justin didn't put the SR into the dicom infrastructure of TCIA.  Erik and I gave him the zip files and I guess he thought it was more organized to keep everything together?

fedorov:
I think we should figure those out before prioritizing! They have manifests to keep things organized, so I don't think that explains it.

DICOMPlugins/DICOMParametricMapPlugin.py
lassoan: The original version `logging.debug('in load(): uid = ', uid)` was not valid syntax.

The updated `logging.debug('in load(): uid = %s', uid)` is correct and fast.

Using f-string is correct, more readable, and slow, as string formatting and all expressions that are used in the string are always evaluated at runtime, regardless the log message is filtered out or not. See for example here https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/).

In this case the performance does not matter, but for consistency and to avoid the risk of slow code getting into performance-critical places by copy-pasting, I would not use f-string for log message formatting.

pieper: Yes, I see, it's `logging` not `print` so the arguments are different.  The correction is fine then 👍

PR #255, opened 2020-11-06
BUG: Fix error reporting in segmentation and TID1500 plugins

lassoan:
DICOMSegmentationPlugin did not report any error when the conversion failed (it raised an exception, but the Python wrapper ignores exceptions).

Also fixed capturing of the exception message: in Python3, exception does not have message attribute anymore.


pieper (approved):
LGTM 👍

fedorov:
Thank you Andras!

PR #259, opened 2021-05-28
COMP: Support bundling in custom application

jcfr:
Since the "<extension_name>Config.cmake" module is not available when
bundling extension sources, this commit excludes the corresponding
requirements.

It also skips the setting of DEPENDENCIES_ADDITIONAL_MODULE_PATHS
because modules associated with the SlicerDevelopmentToolbox, DCMQI
and PETDICOMExtension extensions are built along side the QuantitativeReporting
ones.


jcfr:
Thanks @fedorov :pray:

PR #260, opened 2021-05-28
ENH: Add AI result SR support

pieper:
This parses specific TID 1500 AI results that contain a rectangle and a score for each detection.

It is fairly special-purpose but make this code a more general than the previous version
that only handled lines.  Longer term a more general approach will be needed, probably
by incorporating [highdicom](https://github.com/MGHComputationalPathology/highdicom) as a Slicer dependency.

### DIFF
--- a/DICOMPlugins/DICOMTID1500Plugin.py
+++ b/DICOMPlugins/DICOMTID1500Plugin.py
@@ -20,6 +20,7 @@ class DICOMTID1500PluginClass(DICOMPluginBase, ModuleLogicMixin):
 
   UID_EnhancedSRStorage = "1.2.840.10008.5.1.4.1.1.88.22"
   UID_ComprehensiveSRStorage = "1.2.840.10008.5.1.4.1.1.88.33"
+  UID_Comprehensive3DSRStorage = "1.2.840.10008.5.1.4.1.1.88.34"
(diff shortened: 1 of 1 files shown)


lassoan:
It is great that you use markups line (and not rulers anymore)!

Are you creating a model node for ROIs? We have not markups ROI, which should work much better. It supports automatic and custom (static) measurements with proper DICOM units and terminology support, so with that you can do lossless roundtrip of DICOM import/export, etc. Let me know if you have any questions, or we can discuss at the next developer meeting.

pieper:
Yes, we would prefer markup ROIs but there were 1029 boxes in the SR I was working with and that seemed unworkable until [this issue](https://github.com/Slicer/Slicer/issues/5618) is resolved.  This is all due for a revamp once we start using highdicom instead of all this manual parsing.

lassoan:
As a quick workaround, we could turn off interaction and always (or never) show labels, so that we don't need to keep querying the z buffer. I've asked @Sunderlandkyl to investigate. It would be really bad if after all the invested efforts, markups could not be used because of this performance problem.

fedorov:
@Sunderlandkyl any updates?

fedorov:
@pieper does it make sense to consider support of those planar annotation types in this PR as well: https://github.com/OHIF/Viewers/issues/1215#issuecomment-854237685

PR #264, opened 2022-03-10
BUG: Fix duplicate segments appearing when importing single-slice segmentations

lassoan:
Single-slice segmentations were loaded as an image stack (from files 1.nrrd, 2.nrrd, ...), therefore each volume contained every label.
Fixed by enforcing loading each nrrd file as a single file.

Fixes the problem described at https://discourse.slicer.org/t/incorrect-number-of-segments-when-loading-dicom-segmentation-file/22387


fedorov:
@lassoan the same version of slicer that fails to load correctly the sample in the thread above works with another sample, which also has single-slice segmentations (that sample is not public, I shared it with you via dropbox privately - hope you received it). The segmentations in that sample were created directly using dcmqi, not using Segment Editor. Also, with that sample, there are no errors about "Entire body" in the error log. Do you have any comments about this?

Adding @deepakri201, since she is the one who identified the issue originally.

lassoan:
Thanks, I've received your files. Since the segmentation files dimensions are 512x512x54, ITK loads each file as a single file and does not attempt to load them as a stack of frames. That's why everything works well without specifying singleFile option.

The `GetRegionInAnatomicContext: Failed to find region 'Entire Body' in anatomic context 'Segmentation'` warning is logged because `AnatomicRegionSequence` was not specified when the user set the segment terminology in the Segment Editor. I'll update the importer plugin to not set "entire body" as anatomic context by default but leave it undefined if it was not defined in the DICOM segmentation object.

lassoan:
I've pushed two more commits into this branch:
- leave type, category, anatomic region undefined if they were not defined in the segmentation (this fixes the anatomic context warning that we discussed above)
- better preserve segment label if user changes the terminology (I've just come across as testing things - it was too easy to accidentally overwrite the segment label when modifying the terminology)

fedorov:
Thank you Andras!

@deepakri201 could you possibly test these fixes?

lassoan:
Great, thank you for testing.

PR #266, opened 2022-11-25
Fixes deletion of nested tempdir

marcus-wirtz-snkeos:
By default `slicer.util.tempDirectory()` already contains the current timestamp - the additional subdirectory `self.currentDateTime` in `self.tempdir` leaves you over time with a collection of empty `slicer.util.tempDirectory()` (cf. image). Despite they are in a temp directory, I think it is avoidable ;-)

image


fedorov:
Nice catch, thank you!

Issue #267, opened 2022-11-25
Unexplicit error for export DICOMSegmentation from segmentation node when terminology information not set


marcus-wirtz-snkeos:
If the terminology entry of a segment is unset, the following line updates an empty dictionary:

https://github.com/QIICR/QuantitativeReporting/blob/d875b51db816d28ed729123bf281079272ed3e42/DICOMPlugins/DICOMSegmentationPlugin.py#L663

In the following execution of `itkimage2segimage`, this leads to a really cryptic error message (`cliNode.GetErrorText()` is empty)
```
Traceback (most recent call last):
  File "C:\Users\<user>\AppData\Local\NA-MIC\Slicer 5.0.3\NA-MIC\Extensions-30893\QuantitativeReporting\lib\Slicer-5.0\qt-scripted-modules\DICOMSegmentationPlugin.py", line 367, in export
    exporter.export(exportable.directory, segFileName, metadata)
  File "C:\Users\<user>\AppData\Local\NA-MIC\Slicer 5.0.3\NA-MIC\Extensions-30893\QuantitativeReporting\lib\Slicer-5.0\qt-scripted-modules\DICOMSegmentationPlugin.py", line 569, in export
    raise RuntimeError("%s:\n\n%s" % (cliNode.GetStatusString(), cliNode.GetErrorText()))
RuntimeError: Completed with errors:
```

and in Windows to the pop-up:
<img src="https://user-images.githubusercontent.com/24655255/204051528-dde1d4ba-cf00-4ea0-ae75-fcf1547a6105.png" width="400">

I can't really judge if this might be a particular issue of my setup, but if the DICOM SEG export only works smoothly with terminology information set, then it might be better to add at least a debug log, e.g. here:
https://github.com/QIICR/QuantitativeReporting/blob/d875b51db816d28ed729123bf281079272ed3e42/DICOMPlugins/DICOMSegmentationPlugin.py#L686

I am open to propose a PR.

fedorov:
I agree, adding a check you mentioned makes sense. A PR is welcomed!

PR #268, opened 2022-11-29
Add warning if terminology information is missing in segment (fixes issue #267)

marcus-wirtz-snkeos:
Fixes issue #267 

Additionally, remove unnessecary mkdir() call since directory is created already by Slicer.util


DICOMPlugins/DICOMSegmentationPlugin.py
```diff
@@ -462,7 +462,6 @@ def __init__(self, segmentationNode, contentCreatorName=None):
     self.contentCreatorName = contentCreatorName if contentCreatorName else "Slicer"
 
     self.tempDir = slicer.util.tempDirectory()
-    os.mkdir(self.tempDir)
```
marcus-wirtz-snkeos, line 465: This is a bugfix from my previous PR - this will fail in the previous version since `slicer.util.tempDirectory()` already creates the directory. Apologies for not testing this before...

PR #269, opened 2023-01-02
ENH: add bidimensional measurement support

pieper:
Before, only single length types were supported (SRT G-D7FE).

This commit adds support for measurements with a long axis (SRT G-A185) and a short axis (SRG G-A186) by creating one markup line node for each axis of each measurement.

This support was added for the LNQ project, where we have ~8,700 annotated lymph nodes in ~3,200 series in SR format.

### DIFF
--- a/DICOMPlugins/DICOMTID1500Plugin.py
+++ b/DICOMPlugins/DICOMTID1500Plugin.py
@@ -36,6 +36,9 @@ class DICOMTID1500PluginClass(DICOMPluginBase, ModuleLogicMixin):
       "trackingUniqueIdentifier": { "scheme": "DCM", "value": "112040" },
       "findingSite": { "scheme": "SRT", "value": "G-C0E3" },
       "length": { "scheme": "SRT", "value": "G-D7FE" },
+      "longAxis": { "scheme": "SRT", "value": "G-A185" },
+      "shortAxis": { "scheme": "SRT", "value": "G-A186" },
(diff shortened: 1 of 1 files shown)


pieper:
Here's an example of an SR with two bidimensional measurements.  We could consider making a custom markup class for this type of measurement, but for now it's two lines.  Also the name of the markup is the PersonName of the annotator, which is unknown in this data.  Might be better to name based on the FindingSite and differentiate the long and short axes (or only apply the name to one of them).

image

fedorov:
Steve, thank you for the contribution. Can you share a sample? Did you validate it?

pieper:
No, I don't have a sample I can share.

I'm not sure what you mean by validate.  LIke dciodvfy?

fedorov:
Can you make a sample on a public dataset? 

For validation of TID 1500 SRs, there is a specialized tool `DicomSRValidator` in PixelMed Toolkit (there is a note about it here: https://qiicr.gitbook.io/dcmqi-guide/opening/cmd_tools/sr/tid1500writer).

pieper:
I can't share one of my current examples, but I've tested several and they work with this PR.

Regarding validation, I tried adding index.html as @dclunie suggested, but I don't see the validator `DicomSRValidator.sh` described in the dcmqi guide.

https://www.dclunie.com/pixelmed/software/index.html

Is there another place to look?

fedorov:
Steve, it is in this package pixelmedjavadicom_binaryrelease.20221004.tar.bz2, which is available under this folder https://www.dclunie.com/pixelmed/software/20221004_current/index.html.

PR #270, opened 2023-02-20
allow import of multiple segments

fedorov:
dependent on https://github.com/QIICR/dcmqi/pull/464


fedorov:
Merging, since the dcmqi dependency is now merged as well!

PR #274, opened 2023-11-14
M3D DICOM Modality plugin code

ccosmin97:
Hi, I have added in this pull request the python scripts taking care of the conversion of DICOM M3D objects to Slicer ModelNodes.

Future fixes/features to potentially add can be the ability to convert the ModelNode to a SegmentationNode or LabelmapNode. Here are some pointers on how to do this, thanks to @pieper :

> Instead you if you want your plugin to create a segmentation automatically as part of the DICOM load process you need to call the [logic method](https://github.com/Slicer/Slicer/blob/5f2e83cf71db4bbbe19b957c3b9c01bb7f067267/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx#L1268). That is, convert this code to python:
> 
> https://github.com/Slicer/Slicer/blob/5f2e83cf71db4bbbe19b957c3b9c01bb7f067267/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx#L1189-L1202

The main plugin script is DICOMPlugins/DICOMM3DPlugin.py. The base plugin code, where we retrieve the necessary DICOM metadata for creating the ModelNode is located in DICOMPlugins/base/DICOMM3DPluginBase.py.

@fedorov


DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+
+  def examineFiles(self,files):
+    """ Returns a list of DICOMLoadable instances
+    corresponding to ways of interpreting the
+    files parameter.
+    """
+    loadables = []
+
+    # just read the modality type; need to go to reporting logic, since DCMTK
+    #   is not wrapped ...
```
pieper, line 38: Try to make comments self-contained so that they will be meaningful to other developers in the future.  In this case, if the comment is needed at all, include some extra context about why the wrapping of DCMTK is important to mention here.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+
+    for cFile in files:
```
pieper, line 40: Style nitpick: we avoid using single letters to differentiate variable names.  Here, what does the 'c' mean?  Maybe you mean `candidateFile`?  If so that would be a good name.  With modern editors and auto completion this shouldn't make programming any harder but it makes reading the code easier.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+      uid = slicer.dicomDatabase.fileValue(cFile, self.tags['instanceUID'])
+      if uid == '':
+        return []
+
+      desc = slicer.dicomDatabase.fileValue(cFile, self.tags['seriesDescription'])
+      if desc == '':
+        desc = "Unknown"
+      
+      # EncapsulatedDocument = slicer.dicomDatabase.fileValue(cFile, self.tags['EncapsulatedDocument'])
+      # EncapsulatedDocumentLength = slicer.dicomDatabase.fileValue(cFile, self.tags['EncapsulatedDocumentLength'])
```
pieper, line 51: More style suggestions: we avoid leaving in 'dead code' unless it's something that future developers might need for debugging, and if so an explicit comment describing that use is good practice.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+
+      isDicomM3D = (slicer.dicomDatabase.fileValue(cFile, self.tags['modality']) == 'M3D')
+
+      # print(isDicomM3D)
+      # print(desc)
+      # print(self.tags['modality'])
+      # print(cFile)
+      # print(uid)
```
pieper, line 59: Same here about dead code.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+      return False
+
+    self.tempDir = os.path.join(slicer.app.temporaryPath, 
+                                "QIICR", "M3D",
+                                  self.currentDateTime, loadable.uid)
+    print(self.tempDir)
+    try:
+        os.makedirs(self.tempDir)
+    except OSError:
+      pass
```
pieper, line 108: Here you could consider using `slicer.util.tempDirectory` for brevity and to ease future maintenance.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+    
+    # produces output label map files, one per segment, and information files with
+    # the terminology information for each segment
+    STLFileName = slicer.dicomDatabase.fileForInstance(uid)
+    if STLFileName is None:
+      logging.error('Failed to get the filename from the DICOM database for ' + uid)
```
pieper, line 114: Another nitpick: we try to avoid using capital letters at the start of local variables.  So here you could use `stlFileName`.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+      self.cleanup()
+      return False
+    
+    read_buffer = io.BytesIO(self.GetEncapsulatedDocument(loadable))#bytes(loadable.EncapsulatedDocument))
+    buffer_view = read_buffer.getbuffer()
+    if (int(self.GetEncapsulatedDocumentLength(loadable)) % 2) != 0:
+      buffer_view = buffer_view[0:(read_buffer.getbuffer().nbytes -1)]
+    else:
+      pass
```
pieper, line 123: Not needed, right?

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+    #validity checks !!!!!
+    ####
+    self._createModelNode(loadable, STLFilePath)
+    
+    self.cleanup()
+    
+    return True
+
+  def _createModelNode(self, loadable, STLFilePath):
+    ModelNode = slicer.util.loadModel(STLFilePath)#slicer.mrmlScene.AddNewNodeByClass("vtkMRMLModelNode").AddModel(STLFileName)
```
pieper, line 140: here again, avoid capitalized local variable name and remove dead code.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,174 @@
...
+  
+class DICOMM3DPlugin:
+  """
+  This class is the 'hook' for slicer to detect and recognize the plugin
+  as a loadable scripted module
+  """
+  def __init__(self, parent):
+    parent.title = "DICOM M3D Object Import Plugin"
+    parent.categories = ["Developer Tools.DICOM Plugins"]
+    parent.contributors = ["Andrey Fedorov, BWH"]
```
pieper, line 156: Update this to give yourself credit Cosmin : D

DICOMPlugins/base/DICOMM3DPluginBase.py
```diff
@@ -0,0 +1,92 @@
+import slicer
+import pydicom
+import logging
+from datetime import datetime
+from DICOMLib import DICOMPlugin
+import shutil
+
+
+class DICOMM3DPluginBase(DICOMPlugin):
```
pieper, line 9: I understood this code is already in QuantitativeReporting so can you re-use that?

pieper (changes requested):
Hi @ccosmin97 generally this looks great, but a added a bunch of style comments so this will be more consistent with the rest of the Slicer codebase.

Also as we discussed I think converting to a segmentation makes sense for this datatype.

DICOMPlugins/DICOMM3DPlugin.py
fedorov: Please!

fedorov:
> Future fixes/features to potentially add can be the ability to convert the ModelNode to a SegmentationNode or LabelmapNode. 

I think ability to convert to LabelmapNode is definitely out of scope. This can and should be done using existing generic functionality in Slicer.

But based on the discussion yesterday, I do think that it is best to load this STL into a Segmentation node and not Surface model node, since we have all the accompanying DICOM metadata to properly fit it into the hierarchy. And also since we are introducing this functionality into dcmqi, and it will be consistent with handling SEG.

@lassoan this PR might be of interest to you as well. DICOM M3D encapsulates STL.

DICOMPlugins/base/DICOMM3DPluginBase.py
ccosmin97: yes this .py script is a duplicate of QuantitativeReporting/DICOMPluginBase.py, with a few added attributes, specific to M3D DICOM, i did not what was correct here, have everything in the same DICOMPlugin class or separate ones for M3D and SEG DICOM.

pieper: My feeling is that there's not enough to justify another base class here.  There are not likely to be any other classes that derive from this `DICOMM3DPluginBase` so to me it makes more sense to put anything specific to M3D into the plugin class itself.

fedorov: I agree!

fedorov:
@ccosmin97 as discussed today, let's also try to add loading of the referenced series as an option (as in the DICOM SEG plugin), if such reference is available.

ccosmin97 (commented):
I applied all reviews for the PR, thank you @Steve!

Right now, we still load a ModelNode into the scene from the M3D DICOM object, next step is switching to a SegmentationNode directly instead of the ModelNode.

ccosmin97:
As discussed, I changed the import of the DICOM M3D object from ModelNode to SegmentationNode.
What is left is to properly setup the image hierarchy series, as usually done for SEG DICOM objects. The difference here is M3D objects, at least based on my knowledge, do not have RefSerieUID encoded, we could use FrameOfReferenceUID instead.

lassoan:
There are several ways to to reference the source image in the [Encapsulated STL IOD](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_a.85.html), but I'm not sure which one is the most appropriate.

For example, there is [`Source Instance Sequence`](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.24.2.html#table_C.24-2) in `Encapsulated Document Module`. But there is also [`Common Instance Reference`](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.12.2.html).

I would check what references are used in encapsulated STL data sets in the wild; and also ask @dclunie about which ones are supposed to be used. Hopefully they are the same.

fedorov:
@lassoan the only encapsulated STL that I am aware of are those shared in https://zenodo.org/records/10069911. Those objects were created using PixelMed converter from @dclunie, which to my knowledge is the only converter that exists. That converter did not populate the references, and we will have those shared in IDC in the upcoming v17 data release. So we need to make sure that the plugin is able to load this STL when references are not present. Should they be loaded outside of the hierarchy in this case?

lassoan:
> Should they be loaded outside of the hierarchy in this case?

By not having a reference you only lose the ability to automatically offer loading of the source image.

The series should still have full patient/study/series information, so you should still put the imported series in the proper patient/study hierarchy. It is a single Python function call that is used by most DICOM import plugins - let me know if you cannot find it.

ccosmin97 (commented):
Fixed styling issues and comments.

ccosmin97:
@pieper @fedorov @lassoan 

The 'final' commit was pushed to the PR.

Currently, the DICOM M3D object is loaded as a segmentation Node.
It supports conversion to labelmap/Model. 
PatientID/StudyUID are propagated properly into the M3D objects, so the hierarchy looks correct between the image and the loaded segmentation Node from the M3D object.

Please let me know if more changes are needed, and thank you for the help!

Here is a demo:
gif_final

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+        desc = "Unknown"
+
+      #read EncapsulatedDocument for candidateFile
+      docFile = self.getEncapsulatedDocument(candidateFile)
+
+      #read EncapsulatedDocumentLength for candidateFile
+      docLengthFile = self.getEncapsulatedDocumentLength(candidateFile)
+
+      #read modality type to flag M3D object.
+      isDicomM3D = (slicer.dicomDatabase.fileValue(candidateFile, self.tags['modality']) == 'M3D')
```
pieper, line 57: This should go first, so you can quickly bail out if the modality is not M3D.  Otherwise you will be reading every candidate file (which could be a giant series of thousands of files) over and over again!

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+      desc = slicer.dicomDatabase.fileValue(candidateFile, self.tags['seriesDescription'])
+      if desc == '':
+        desc = "Unknown"
+
+      #read EncapsulatedDocument for candidateFile
+      docFile = self.getEncapsulatedDocument(candidateFile)
+
+      #read EncapsulatedDocumentLength for candidateFile
+      docLengthFile = self.getEncapsulatedDocumentLength(candidateFile)
+
```
pieper, line 55: These methods both read the full file, which may take lots of time.  I suggest you refactor and maybe combine these into a single method that does both operations.

But also, do you really need to read the full file just to know if it's an M3D file?  The load step should be as lightweight as possible so that you only invest time reading the data if you are sure it's the right interpretation for the files (e.g. it has the highest confidence or the user has specifically selected it).

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+  def load(self,loadable):
+    """ Load the DICOM M3D object
+    """
+    logging.debug('DICOM M3D load()')
+    try:
+      uid = loadable.uid
+      logging.debug('in load(): uid = '+uid)
+    except AttributeError:
+      return False
```
pieper, line 113: It's better style (shorter, more efficient) to use hasattr rather than a try/except.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+
+    self.tempDir = slicer.util.tempDirectory()
+    print(self.tempDir)
```
pieper, line 116: Remove the debugging code or convert to logging

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+    segmentationNode.SetName(loadable.name)
+
+    # Initialize Segmentation Display Node
+    segmentationDisplayNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLSegmentationDisplayNode")
+    segmentationNode.SetAndObserveDisplayNodeID(segmentationDisplayNode.GetID())
+
+    #Setup converters for further export to labelmap or model representations 
+    vtkSegConverter = vtkSegmentationCore.vtkSegmentationConverter
+    segmentation = vtkSegmentationCore.vtkSegmentation()
+    segmentation.SetMasterRepresentationName(vtkSegConverter.GetSegmentationBinaryLabelmapRepresentationName())
```
pieper, line 167: Wouldn't it make more sense for the closed surface model to be the Master representation here, since that's what came in the file.  Converting to a labelmap first is lossy since it relies on a heuristic to determine the labelmap resolution.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+    segmentation.CreateRepresentation(vtkSegConverter.GetSegmentationClosedSurfaceRepresentationName(), True)
```
pieper, line 168: Here you are effectively re-creating the surface model from the labelmap created from the original surface model.  It would be better to keep the original one.

DICOMPlugins/DICOMM3DPlugin.py
```diff
@@ -0,0 +1,641 @@
...
+    dNode = modelNode.GetDisplayNode()
+    if dNode is not None:
+      slicer.mrmlScene.RemoveNode(dNode)
+    slicer.mrmlScene.RemoveNode(modelNode)
+
+  def examineForExport(self, subjectHierarchyItemID):
+    shNode = slicer.vtkMRMLSubjectHierarchyNode.GetSubjectHierarchyNode(slicer.mrmlScene)
+
+    exportable = self._examineExportableForSegmentationNode(shNode, subjectHierarchyItemID)
+    return self._setupExportable(exportable, subjectHierarchyItemID)
```
pieper, line 203: This doesn't appear to export as M3D.  Is this just a duplicate of the code from the segmentation plugin?  If so it should be removed.  I don't remember the exact method but I believe that if your plugin doesn't support exporting you either project a stub of an exporter class or leave it out completely.

pieper (changes requested):
This is looking close, but please have a look at the issues I pointed out.  Let me know if you want to discuss any of them.

DICOMPlugins/DICOMM3DPlugin.py
ccosmin97: > These methods both read the full file, which may take lots of time. I suggest you refactor and maybe combine these into a single method that does both operations.

I combined both of these methods, also they are not necessary to determine if it's an M3D object, only the modality tag is the indicator for M3D.

Instead of creating loadable attributes for EncapsulatedDocument and EncapsulatedDocumentLength inside examineFiles, I call these methods directly into the load function.

fedorov:
@ccosmin97 nice progress!

@pieper thank you very much for the review and guidance!

>> Should they be loaded outside of the hierarchy in this case?
>
> The series should still have full patient/study/series information, so you should still put the imported series in the proper patient/study hierarchy. 

@lassoan of course, you are right! I was writing that comment while in another meeting, and didn't think long enough about this. Sorry!

fedorov:
@ccosmin97 what is the status of this?

ccosmin97 (commented):
Reviewed all changes and added/modified requested styling/functional suggestions.
Thank you @pieper!

ccosmin97:
> @ccosmin97 what is the status of this?

@fedorov Suggestions in the second Steve's review are implemented.

Here is a snapshot of the current implementation, where we load the M3D DICOM object into the database, and convert it to a segmentation node without reference image information. This segmentation node can then be exported to a model node or a labelmap node.

By default, the unique corresponding segment inside the segmentation object is named 'Segment 1', with a default color.
 
final_m3d

fedorov:
I tested, and it looks good. Thank you for your contribution @ccosmin97! 👍

lassoan:
Thank you @ccosmin97, nice work!  It would be great if you could write a post on the Slicer forum (discourse.slicer.org) to announce this new feature.

fedorov:
We missed it during the review, but the new plugin was not added to CMakeLists, and thus was not available in Slicer nightlies. It is hopefully fixed by this commit: https://github.com/QIICR/QuantitativeReporting/commit/55989e4d91b17aa1d4b4ce632362e8a2a4bf0f66.

lassoan:
If everything works then please also send a pull request to update the [list of supported file formats](https://slicer.readthedocs.io/en/latest/user_guide/data_loading_and_saving.html#supported-data-formats). Thank you!

ccosmin97:
@lassoan We have a draft ready for posting an announcement in Slicer forum, but I do not have permission to create a new topic in the Announcement section.

Could you help with this issue? 
Thank you!

lassoan:
All Slicer forum members (8000+ people) are signed up to get email notification about all new topics in the Announcement category, with the promise that we keep the traffic in the category low. If we start to send too many emails we risk that users will unsibscribe from this category or the entire forum. Breakthrough achievements that affect majority of Slicer users can definitely be published in the Announcements category, but for most new features we use the "Support" category and attach the "feature" tag. 

It is hard to draw a line for what is an important feature that warrants a site-wide announcement, but in case of the M3D importer I would lean towards using the regular path (posting in Support category).

@fedorov what do you think?

fedorov:
Andras, thanks for the explanation! I agree, it makes sense to just post to the Support category.

ccosmin97:
Sounds good, that makes sense, thank you @lassoan .

PR #281, opened 2025-03-22
Update DICOMSegmentationPlugin.py for change in name of context

pieper:
https://github.com/Slicer/Slicer/commit/65409219f724bf03bd0307e9c2062c06030f3951


pieper:
Address #282

PR #290, opened 2025-08-25
Replace deprecated pydicom function

marcus-wirtz-snkeos:
Replace deprecated function `pydicom.read_file()` by `pydicom.dcmread()`, cf.:
https://github.com/pydicom/pydicom/pull/506

The extension is currently not functional with preview version 3DSlicer 5.9, since in there python 3.11 will install `pydicom>=3.0` by default. This PR fixes this issue.

### DIFF
--- a/DICOMPlugins/DICOMM3DPlugin.py
+++ b/DICOMPlugins/DICOMM3DPlugin.py
@@ -64,14 +64,14 @@ class DICOMM3DPluginClass(DICOMPluginBase):
 
   def getFrameOfReferenceUID(self, candidateFile):
     """Returns the frame of referenceUID for the given loadable"""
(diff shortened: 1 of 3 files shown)


fedorov:
@marcus-wirtz-snkeos I realized too late that the latest stable is still using python 3.9, and will install an earlier pydicom version. Can you add a try/catch block to make sure the code works with both versions of pydicom?

pieper:
Doing an explicit check for the method is probably cleaner than try/catch.  I.e. `hasattr(pydicom, "read_file")`

fedorov:
> `pydicom.dcmread()` is very old and around since ~2018 (even version 1.2.0 already had it)

I didn't realize that! Then I think we should be good.

